PR #6, opened 2019-06-24
chore: update lockfile

Knorrke:
Update lockfile, fixing build issues of `node-webcrypto-ossl` on windows.


Knorrke:
@inyono Circle CI wasn't working before, was it?

inyono:
Should be fine in master now :)

PR #12, opened 2019-07-12
refactor: build docker images without local dev environment

inyono:
- [x] Update README before merging


packages/private/docker/index.ts
```diff
@@ -48,12 +48,15 @@ export function buildDockerImage({
...
-}
 
-export function publishDockerImage(options: DockerImageOptions) {
-  const { name, version } = options
-  buildDockerImage(options)
+  if (
+    !process.env.CI ||
+    !process.env.CIRCLE_BRANCH ||
+    process.env.CIRCLE_BRANCH !== 'master'
+  )
```
Knorrke: does this mean only CI can publish the docker images? I think you had this different before, that you could publish stuff manually as well, what was the reasoning behind the change?

packages/public/editor-renderer/Dockerfile
```diff
@@ -1,21 +1,11 @@
...
+FROM base as release
+COPY --from=serlo-org-workspace /usr/src/app/packages/private/legacy-editor-to-editor packages/private/legacy-editor-to-editor
+COPY --from=serlo-org-workspace /usr/src/app/packages/private/editor-helpers packages/private/editor-helpers
+COPY --from=serlo-org-workspace /usr/src/app/packages/private/markdown packages/private/markdown
+COPY --from=serlo-org-workspace /usr/src/app/packages/public/editor-renderer packages/public/editor-renderer
+COPY --from=serlo-org-workspace /usr/src/app/node_modules node_modules
 WORKDIR packages/public/editor-renderer
-ENTRYPOINT node .
+ENTRYPOINT ["node", "."]
 EXPOSE 3000
```
Knorrke: Do you need to change this because `EDITOR_RENDER_PORT` is set to 3001?

packages/private/docker/index.ts
inyono: Good point. I changed this up so that we don't have two different jobs on CircleCI building the Docker Image. But yeah, it should be able to depoy images manually, too, gonna check that.

packages/public/editor-renderer/Dockerfile
inyono: To be honest, I don't care about that environment variable in production ^^' It's just there to be able to override it in dev.

inyono:
With this PR, building images works without any local environment (except that we use `yarn` as a task runner). This is mostly so that the production images don't rely on our local environment (that might be different depending on OS). For development, we still need `yarn` for most tasks (e.g. running tests) and the `workflow` part of the docker-compose relies on local node_modules (If really needed/desired, we can install that in the container, too, but I'd rather avoid additional abstractions)

packages/private/docker/index.ts
inyono: Solved. Should work now by setting environment variable `DEPLOY` to true, e.g. `DEPLOY=true yarn build:docker`

PR #22, opened 2019-07-22
implement converter from splish to edtr-io

Knorrke:
:tada:


Knorrke:
There's an issue with `@edtr-io/plugin-important-statement`, doesn't seem to include the build files, only src.

inyono:
TODO @inyono:
- [ ] Backward-compatibility in client & editor-renderer
- [ ] Is always `importFromHtml` a breaking change?

inyono:
Superseded by #27

PR #23, opened 2019-07-22
replace splish editor renderer with editor

Knorrke:
This builds up on #22. Tests already tell that the rendering isn't working, so we have to check that. Its failing because of the Katex css file.


inyono:
Superseded by #27

PR #27, opened 2019-07-30
Replace Splish editor with Edtr.io

inyono:
Supersedes #22 and #23.

### Open tasks
- [x] Integrate Edtr.io in client
- [x] Integrate Edtr.io in server
  - [x] Especially: fake plugins for every content type as discussed
  - [ ] https://github.com/serlo/athene2/pull/884

### Before merging / releasing
- [ ] Test every content type (both legacy to Edtr.io & Splish Editor to Edtr.io)
- [ ] Test pages
- [ ] Test profiles
- [x] If possible, wait for 100% Kubernetes because legacy backend has completely different deployment process for editor-renderer
- [x] (Is always importFromHtml a breaking change?)
- [x] (Check Backward-compatibility)


inyono:
Pushed a couple of fixes, `editor-renderer` fails for some easy Splish editor state, though. Let's look into that together, might be missing something simple...

Knorrke:
`editor-renderer` should be fixed now, also rebased on master

inyono:
Need to explore why `athene2-assets` doesn't build anymore

inyono:
Merged to get this into staging :) Still needs some polishing before going into production, though.

PR #64, opened 2019-10-09
improve content api

### LINKED ISSUE #48: Content-API redirect to parents
When requesting e.g. one `grouped-text-exercise` of a bigger exercise the request will be redirected to the parent. Redirect also happens in the content-API, but we want to be able to embed single `grouped-text-exercise` entites.

The redirects to parent were introduced for some small entity types to prevent annoying edit flows when creating e.g. Sc-Mc-exercises. With the new editor we optimized the workflow in a different way so maybe we don't need the redirects at all?

Knorrke:
fixes some issues with content api (see below) and fixes #48

### Fixes
- hide header of entities without title
- escape component redirects (e.g. subexercise to exercise)
- reduce empty space on small elements
- fix height calculation
- include fontawesome css instead of dynamicly adding it in js


packages/public/client/src/main/styles/base/_layout.scss
```diff
@@ -30,7 +30,7 @@
   }
 }
 
-#page {
+#page:not(.no-indent) {
```
inyono: Alternatively:

```
#page.no-ident {
  margin-top: 0;
}```

File: packages/public/server/src/module/Ui/templates/layout/2-col.phtml [hunk 2]
```diff
@@ -57,7 +60,7 @@ $pageClass = join(' ', $pageClasses);
     <div id="page" class="<?php echo $pageClass ?>">
         <?php echo $this->partial('layout/partials/left-sidebar'); ?>
         <section class="main clearfix">
-            <div id="content-layout" class="content clearfix">
+            <div id="content-layout" class="content clearfix <?php echo $this->layout()->usingAPI ? 'api' : '' ?>">
```
inyono: Set class on body instead (and select with `.api #content-layout`

packages/public/client/package.json
```diff
@@ -30,6 +30,7 @@
     "dnode": "^1.0.0",
     "font-awesome": "^4.0.0",
     "howler": "^2.0.0",
+    "iframe-resizer": "^4.2.1",
```
inyono:
```suggestion
    "iframe-resizer": "^4.0.0",
```

packages/public/server/src/module/Ui/templates/entity/page/grouped-text-exercise.twig
```diff
@@ -23,11 +23,12 @@
 {% set parent = entity.getParents('link').first() %}
 {% do headTitle(title ~ ' - ' ~ entity().findSubject(parent) ~ ' ' ~ (entity.getType().getName() | trans)) %}
 {% do normalize().headMeta(entity) %}
-{{ pageHeader(normalize().toTitle(entity))
-    .setSubtitle(entity.getType().getName() | trans)
-    .setBacklink(normalize().toUrl(entity.getParents('link').first()))
-    .render(false) }}
-
+{% if not layout().usingAPI %}
+    {{ pageHeader(normalize().toTitle(entity))
```
inyono: Maybe: remove page headers completely (since ID isn't a good title) and rather only show the type in Breadcrumbs (or don't show it at all)

yarn.lock
```diff
@@ -13649,6 +13667,14 @@ react-hotkeys@^2.0.0:
   dependencies:
     prop-types "^15.6.1"
 
+react-iframe-resizer@^1.0.1:
```
inyono: `yarn.lock` out of date

PR #77, opened 2019-10-14
add table for user fields and save interests on registration

### LINKED ISSUE #69: Move user information of registration to profiles
- Move "Ich bin hier als..." from newsletter to general user profile and make the field required
- Still send the value to MailChimp (if the user signed up to newsletter)
- ~Make editable in user profile (if possible, also update MailChimp here)~
- Should be implemented in such a way that we can add additional user profile fields later (w/o database change)

Knorrke:
closes #69

### DIFF
--- a/packages/public/server/src/module/Ui/templates/user/user/register.twig
+++ b/packages/public/server/src/module/Ui/templates/user/user/register.twig
@@ -72,6 +72,11 @@
             {{ passwordConfirmErrors }}
         </div>
     </div>
+    <div class="form-group{{ interestsErrors ? ' has-error' }}">
+        {{ formLabel(form.get('interests')) }}
+        {{ formElement(form.get('interests')) }}
+        {{ interestsErrors }}
+    </div>
 </fieldset>
 <fieldset>
     <legend>{{ 'Stay in touch!' | trans }}</legend>
@@ -103,11 +108,6 @@
             </div>
             {{ lastNameErrors }}
         </div>
-    </div>
-    <div class="form-group{{ interestsErrors ? ' has-error' }}">
-        {{ formLabel(form.get('interests')) }}
-        {{ formElement(form.get('interests')) }}
-        {{ interestsErrors }}
     </div>
     <div class="form-group">
         <p>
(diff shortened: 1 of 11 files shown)


packages/public/server/src/module/Ui/templates/user/user/register.twig
```diff
@@ -72,6 +72,11 @@
             {{ passwordConfirmErrors }}
         </div>
     </div>
+    <div class="form-group{{ interestsErrors ? ' has-error' }}">
```
inyono: this line looks weird. Is this fine w/o an else case?

packages/public/server/src/module/User/src/User/Manager/UserManagerInterface.php
```diff
@@ -84,4 +86,15 @@ public function getUserFromAuthenticator();
...
+    /**
+     * Adds or sets user fields to a user.
+     *
+     * <code>
+     * $userManager->setFields($user, ['foo' => 'bar', 'acme' => 'bar']);
+     * </code>
+     * @param UserInterface $user
+     * @param array $data
+     */
+    public function setFields($user, $data);
```
inyono: If I understood the implementation correctly: this is more a `mergeFields`, right?

Knorrke: Yes, that's true. I'll rename it, because I do think it is the behaviour we want. We might need a function for removing a field though.

packages/public/server/src/module/Ui/templates/user/user/register.twig
Knorrke: Hmm, doesn't look correct, but it was like that before... I think I'll just add the else case, but seems to work o.O

Knorrke: Well, actually the whole file is like this, and it's supported since twig 1.12.0 according to https://stackoverflow.com/a/24859759 while we're on 1.20.0 so I think this is fine.

inyono: And ppl say that JS is weird xD

inyono:
Btw: moved some subtasks of #69 into new issues

PR #11, opened 2019-07-10
Feature/tamil

Knorrke:
### Fixed:
Allow UTF-8 in Aliases - fixes #10, closes #9 

### Added:
Enable Hindi tenant - closes #8 

For this to work, some additional steps are needed (see #8):
1. Run the sql script in the database to set up the necessary data for a language tenant
2. Afterwords: (ping @jakobwes) Add a license (/license/manage), subjects (/taxonomy/term/organize-all), and pages (/pages). The configured subjects are `community`, `இலக்கணம்`, `வரலாறு`, and one page should have the url `/serlo`

PR #86, opened 2019-10-18
Filter non letter characters from alias

Knorrke:
Filter non letter characters (respecting unicode) from alias, because e.g. this looks weird: https://de.serlo.org/mathe/geometrie/dreiecke,-vierecke,-kreise-andere-ebene-figuren/kreise-kreisteile/berechnungen-kreis

### DIFF
--- a/packages/public/server/src/module/Common/src/Common/Filter/Slugify.php
+++ b/packages/public/server/src/module/Common/src/Common/Filter/Slugify.php
@@ -32,8 +32,11 @@ class Slugify implements FilterInterface
      */
     protected static function slugify($text)
     {
-        $text = trim($text, " ");
         $text = preg_replace("~ +~u", '-', $text);
+        // replace non letter or digits
+        $text = preg_replace('~[^\\pL\\pM\d-]+~u', '', $text);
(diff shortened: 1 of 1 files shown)


packages/public/server/src/module/Common/src/Common/Filter/Slugify.php
```diff
@@ -32,8 +32,11 @@ class Slugify implements FilterInterface
      */
     protected static function slugify($text)
     {
-        $text = trim($text, " ");
         $text = preg_replace("~ +~u", '-', $text);
+        // replace non letter or digits
+        $text = preg_replace('~[^\\pL\\pM\d-]+~u', '', $text);
```
Knorrke: We had the `\p{L}` filter before and removed it in #11, because it filtered too much especially in hindi and tamil names. This is now handled differently with `\p{M}`: `\p{L}` matches a character point and `\p{M}` matches accents combined to a letter, e.g. `à` is build from `\p{L}`matching `a` and `\p{M}` matching the accent (see https://www.regular-expressions.info/unicode.html).

PR #87, opened 2019-10-21
server: update translation files

inyono:
Removes unnecessary strings introduced in the last update of translation files since Poedit fixed their Twig extractor


packages/public/server/src/lang/athene2.po
```diff
@@ -1559,33 +1522,6 @@ msgstr ""
...
-msgid "content"
-msgstr ""
-
-#: ../module/Ui/templates/discussion/partials/form-renderer.twig:29
-msgid "start"
-msgstr ""
-
-#: ../module/Ui/templates/editor/partials/video-modal.twig:35
-msgid "Don't show again"
-msgstr ""
```
Knorrke: Check this please @inyono

packages/public/server/src/lang/athene2.po
```diff
@@ -2530,10 +2357,6 @@ msgstr ""
 msgid "Search results"
 msgstr ""
 
-#: ../module/Ui/templates/static/de/datenschutz/archive.twig:22
-msgid "<li><span>Aktualisierungen: Datenschutzerklärung</span></li>"
-msgstr ""
```
Knorrke: check this please

Knorrke: this is correct, shouldn't be translated

PR #106, opened 2019-11-15
Cloudflare Worker for donation campaign

inyono:
- See https://serlo-donation-campaign.serlo.workers.dev/ for an example of the deployed Cloudflare Worker
- Behavior:
  - Fetches the data from https://docs.google.com/spreadsheets/d/18ri3s1cqw1-jYjnPbPkI_9v4t9wBY1Sp3YtklXO0_Wg and caches it for one hour
  - Chooses the active experiment (i.e. the last experiment with a start date that is not in the future)
  - Assigns the user to a group according to the cookie
  - Responds with the data that we need in the front end


inyono:
Required to finish the stuff in #70

PR #107, opened 2019-11-19
Improvements for e2e testing

kulla:
While working on my first issues on this repo I came across the following little improvements in our e2e testing:

* Files inside `__tests-e2e__` weren't formatted by `yarn format:prettier`
* `jest.setTimeout()` takes the time in milliseconds -> variables needed to be changed


package.json
```diff
@@ -42,7 +42,7 @@
     "start:server": "docker-compose up --build --detach",
     "stop:server": "docker-compose stop",
     "down:server": "docker-compose down",
```
inyono: Nice catch

packages/public/server/__tests-e2e__/home.ts
```diff
@@ -21,8 +21,8 @@
  */
 import { Browser, launch, Page } from 'puppeteer'
 
-const seconds = 1000
-jest.setTimeout(60 * seconds)
```
inyono: This was intended. The second line should be read as: "the timeout is 60 seconds"

kulla: I read the line as "Timeout is 60 seconds" because the variable was named `seconds`. Then I would rename the variable as `milisecondsPerSecond` or I would write `jest.setTimeout(60 * 1000)`...

inyono: > I read the line as "Timeout is 60 seconds" 
Yep, that is what I wrote? :) Or did you mean "Timeout is 1000 seconds"?

If so and you don't accept my explanation (😅), you could move that into a function to make it clearer, e.g.
```
setTimeout(60)

function setTimeout(seconds: number) {
  jest.setTimeout(seconds * 1000)
}
```
which would IMHO combine both of our understandings of the two approaches we seen.

kulla: Okay, I moved it to a global `jest.setup-e2e.js` file in https://github.com/serlo/serlo.org/blob/b0dffdc027016b4798d32dc7549402c855c423ee/jest.setup-e2e.js#L1-L5

inyono: Awesome, thanks

Issue #108, opened 2019-11-19
Remove legacy HTML attributes


kulla:
The HTML attributes `data-id`, `data-edit-type` and `data-edit-field` aren't used any more and thus can be deleted.

inyono:
Closed by #109

PR #109, opened 2019-11-19
Bugfixes for the content api

kulla:
This PR includes fixes fo #90 and #91 (while working on this issue I also removed legacy HTML attributes as described in #108)


packages/public/server/__tests-e2e__/entities/grouped-exercise.ts
```diff
@@ -0,0 +1,63 @@
...
+      expect(await page.$$(pageHeaderSelector)).toHaveLength(1)
+  })
+
+  test.each(['contentOnly', 'hideBanner', 'fullWidth'])(
+    'grouped exercise has no page header when %p is set (content-api)',
+    async (contentApiParam) => {
+      await page.goto(groupedExerciseUrl + '?' + contentApiParam)
+      expect(await page.$(pageHeaderSelector)).toBeNull()
+    }
+  )
```
inyono: IMHO: You should be careful with `test.each` and similar constructs. Avoiding code duplication is no goal in itself (i.e. less code is not necessarily better and sometimes harmful) but rather a mean for other goals, e.g. only having to change one location, etc. Especially in tests, code readability is IMHO way more important than code duplication. It's fine in this case, but combined with `R.xprod` like somewhere below it might get tricky.

inyono: See e.g. https://kentcdodds.com/blog/aha-programming for some reasoning behind that

Issue #116, opened 2019-11-21
CSRF error when saving static pages (new editor)


kulla:
While updating serlo.org/jobs Simon got today the following error:

![image](https://user-images.githubusercontent.com/1327215/69375687-34057280-0ca9-11ea-83d4-67923c723e26.png)

Error: `{csrf: {notSame: "The form submitted did not originate from expected site" }}`

The variable `window.csrf`was set successfully in the browser and it was sent to the server in the payload of the save request.

inyono:
Things to check
- Is the CSRF token stable across different pods? e.g. when pod A renders the "form" (i.e. passes down the token via CSRF) but pod B gets the response, is the CSRF token the same? (pretty sure that is the case since AFAIK the Zend CSRF token is session-specific)
- The Zend CSRF token has a default timeout of 5 minutes. We should check whether we've already overridden the default. Not sure what the best practice is? Refreshing CSRF tokens? Longer timeout? If we still use the default value, we definitely need to do something since 5 mins is not enough for the editor.

inyono:
Removed `edtr-io` label since I'm 99%-sure that the issue isn't in Edtr.io (we more or less pass `window.csrf` through) but rather in the token itself

inyono:
Also: might have been handled in the old editor by default since a failing form would just have rerendered the form with the changes made (and maybe some obscure error message) but you could just resend it again (since rerendering refreshes the token).

inyono:
Yep... The token is the problem. Sometimes getting different tokens in different tabs and they seem to expire. Might be related to our https://sentry.io/organizations/serlo/issues/1148109109/events/?project=1518832 issues...

inyono:
Closed by #118

PR #118, opened 2019-11-25
Refactor csrf token handling using Double Submit Cookie pattern

inyono:
Refactors our csrf token handling using the [Double Submit Cookie pattern](https://cheatsheetseries.owasp.org/cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.html#double-submit-cookie):
- No longer pass "csrf token" down via `window.csrf` (because cashing)
- Instead, pass down a new cookie "CSRF" that contains a pseudorandom value associated with the session (same expiration, same store)
- Validator and form field no longer depend on Zend's default CSRF behavior and have been moved into their own module
- Still: need to prevent caching of pages that contain a form for anonymous users


inyono:
Should fix #116

packages/private/edtr-io/src/plugins/image.tsx
```diff
@@ -90,52 +91,57 @@ export const validateFile: UploadValidator<FileError[]> = file => {
...
+        if (!e.target) return
+        const formData = new FormData()
+        formData.append('attachment[file]', file)
+        formData.append('type', 'file')
+        formData.append('csrf', getCsrfToken())
+
+        axios
+          .post('/attachment/upload', formData)
+          .then(({ data }) => {
+            console.log(data)
```
Knorrke: remove this probably?

inyono:
@inyono: check cashing

inyono:
Okay, I have _no_ idea where we specify that `auth/login` shouldn't be cached... I'd say we ignore the potential issue for now (since the login page isn't cached anyways) and deal with it if it's really a problem (e.g. by inserting the value dynamically via Cloudflare Worker / JS / ...)

PR #120, opened 2019-11-28
converter for solution-step plugin

metzgaria:
converted solutions use the solution-step plugin instead of the rows plugin.

can be merged after solution-step plugin in edtr.io is ready

### DIFF
--- a/packages/private/edtr-io/src/deserialize.ts
+++ b/packages/private/edtr-io/src/deserialize.ts
@@ -542,9 +542,19 @@ export function deserialize({
       changes: '',
       // FIXME: solutions don't have a title
       title: '',
-      content: serializeEditorState(
-        toEdtr(deserializeEditorState(state.content))
-      )
+      content: serializeEditorState({
+        plugin: 'solutionStep',
+        state: {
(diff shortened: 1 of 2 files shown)


packages/private/edtr-io/src/deserialize.ts
```diff
@@ -542,9 +542,19 @@ export function deserialize({
...
-      )
+      content: serializeEditorState({
+        plugin: 'solutionStep',
+        state: {
+          introduction: { plugin: 'text' },
+          solutionSteps: [
+            {
+              type: 'step',
+              isHalf: false,
+              content: toEdtr(deserializeEditorState(state.content))
```
inyono: Here you still need to catch the case when `state.content` is already a `solutionStep` plugin

inyono:
@metzgaria any updates? :)

metzgaria:
I thought this should be done via state migrator?

inyono:
@metzgaria Nope, you added the code at the right place. But you still need to cover some edge cases (like I commented above). We can discuss that in detail if my comment is too concise.

PR #129, opened 2019-12-03
content-api: remove content types

### LINKED ISSUE #89: content-api: remove content types

kulla:
This PR fixes #89. I also have added a script so that we can easily update the mysql dump in this repository `yarn dump:sql`. Since the old dump was created with `PHPMyAdmin` and the new script uses `mysqldump` there is a big diff between both files.


scripts/dump_mysql_serlo_database.sh
```diff
@@ -0,0 +1,25 @@
...
+#   reduces the noise in diffs between dumps. The dumped date is implicitly
+#   stored by git.
+# * The options --complete-insert and --comments make the dump more readable for
+#   humans.
+docker exec serloorg_mysql_1 /usr/bin/mysqldump --user=root --password=secret \
+	--lock-all-tables --complete-insert --comments --skip-dump-date \
+	--databases serlo \
+		| sed 's$VALUES ($VALUES\n($g' \
+		| sed 's$),($),\n($g' \
+		> "${SQL_INIT_FILE}"
```
inyono: Please don't use shell scripts here since they don't work in Windows. Could you rewrite that in typescript for example? (see `scripts/license-headers.ts` for example)

scripts/dump_mysql_serlo_database.sh
```diff
@@ -0,0 +1,25 @@
...
+
+# * Before and after each row of each INSERT command in the final .sql file a
+#   new line is added with the `sed` commands. Thus we have smaller diffs in
+#   git when table rows are modified (see https://stackoverflow.com/a/19961480).
+# * The option --skip-dump-date omits dumping the current date which further
+#   reduces the noise in diffs between dumps. The dumped date is implicitly
+#   stored by git.
+# * The options --complete-insert and --comments make the dump more readable for
+#   humans.
+docker exec serloorg_mysql_1 /usr/bin/mysqldump --user=root --password=secret \
```
inyono: Please use `docker-compose exec mysql` instead since the actual container names aren't the necessarily the same on different machines

inyono: Doesn't `mysqldump` work, too?

packages/public/server/src/module/Ui/templates/entity/view/event.twig
```diff
@@ -22,7 +22,7 @@
 {% set title = 'event' | trans %}
 {% set revision = entity.getCurrentRevision() %}
 {% set content = renderer().toHtml(revision.get('content')) %}
-<article>
+<article id="description">
```
inyono: why is this `id` needed?

packages/public/server/src/module/Ui/templates/entity/page/math-puzzle.twig
```diff
@@ -24,9 +24,7 @@
 {% do normalize().headMeta(entity) %}
 {% include 'entity/page/partials/placeholders' %}
 <div class="page-header">
-    <h1>{{ title }}
-        <small>{% trans %} Math puzzle {% endtrans %}</small>
-    </h1>
+    <h1 id="title">{{ title }}</h1>
```
inyono: why is this `id` needed?

packages/public/server/src/module/Ui/templates/entity/page/event.twig
```diff
@@ -24,9 +24,7 @@
 {% do normalize().headMeta(entity) %}
 {% include 'entity/page/partials/placeholders' %}
 <div class="page-header">
-    <h1>{{ title }}
-        <small>{% trans %} Event {% endtrans %}</small>
-    </h1>
+    <h1 id="title">{{ title }}</h1>
```
inyono: why is this `id` needed?

scripts/dump_mysql_serlo_database.sh
inyono: Also: did you check whether the new `001-init.sql` still works?

kulla: Yes, I'll convert this to typescript. You can easily test it by `yarn test:e2e` since the new tests uses new entries I have added with the commit https://github.com/serlo/serlo.org/pull/129/commits/60bb0c3b5663b43053265b0a87a94276df051092

scripts/dump_mysql_serlo_database.sh
kulla: Yeah, should :-)

packages/public/server/src/module/Ui/templates/entity/view/event.twig
kulla: For e2e testing. Let's discuss it in https://github.com/serlo/serlo.org/issues/147 and I'll change it afterwards (maybe in a new PR since I also used those testing ID in a last PR).

inyono: I see. Feel free to check my answer in #147 and let me know if that's okay for you / you want to discuss that.

kulla:
Note: My bash script replaced all occurrences of `),(` with `),\n(` - even those which are in the database. Since there are only 2-3 articles where this happened and this `.sql` is only for developing I haven't fixed it (but I can do it easily, when neccesary).

packages/public/server/src/module/Ui/templates/entity/view/event.twig
inyono: This is not addresses yet, is it? At least it didn't match to what we discussed, if I'm not mistaken. Currently the testing approach is tied tightly to the DOM and even needs changes to production code. Instead of
```.js
const title = await page.evaluate(() => {
  return document.querySelector<HTMLElement>('h1#title')!.innerText
})
expect(title).toBe('Beispielveranstaltung')
```
I'd prefer if we rather do something like
```.js
const { getDocument, queries, wait } = require('pptr-testing-library')

const { getByText } = queries

const $document = await getDocument(page)
const $title = await getByText($document, 'Beispieltext')
// If you also want to assert that it is a heading
console.log($title)
// Something like `expect($title.tagName).toEqual('h1')`
```
Similarly for the other stuff. Then, we don't need changes to our production code and our tests are much more resilient against changes in the DOM (e.g. if we render stuff differently)

kulla: Oh, I thought I have added a comment here... You would address this in a new PR where I address all current e2e testing refactoring issues. What is better for you? Shall I address the e2e issues in this PR or shall I start a new one, so that this one can be merged (we may create an issue for tracking the remaining tasks)?

inyono: Follow up issue: #172

scripts/dump_mysql_serlo_database.ts
```diff
@@ -0,0 +1,200 @@
...
+    'interface can be insecure.\n'
+
+  transformLine(line: string): void {
+    if (line !== this.uneccessaryError) {
+      this.push(line)
+    }
+  }
+}
+
+const mysqldump = spawn('docker-compose', dockerComposeArgs)
```
inyono: Something I like to do when coding longer functions / scripts: I usually try to increase the abstraction level when reading the file from top to bottom. This way, you don't necessarily have to read the whole source file but can stop at the appropriate level (resp. jump only down to the lower level abstractions you need to make your changes and can ignore most other stuff). So for example, in this case I'd roughly order the stuff the following way:

```.ts
// All imports (since they have to be top)

// Any global variables I might need. These must be before their first usage. If I need variables only in some lower level abstraction, move them in the function

// I call exec on top-level before its declaration so that it's clear that this is a sc
exec()

// Define exec (async if this returns a promise)
async function exec() {
  const dockerComposeArgs
  const mysqldump = spawn('docker-compose', dockerComposeArgs)

  mysqldump.stdout
    .pipe(new ConcatenateInsertCommands(encoding, maxInsertCmdLength))
    .pipe(fs.createWriteStream(sqlInitFile))

  mysqldump.stderr
    .pipe(new IgnoreInsecurePasswordWarning('utf8'))
    .pipe(process.stderr)

  mysqldump.on('error', error => {
    console.error('ERROR: ' + error)
  })
  mysqldump.on('exit', (code, signal) => {
    process.exit(code !== null ? code : 1)
  })
}

// Now come all the needed classes / functions, might be even part of the function body itself
// Also sorted in the same way. I.e. on the top are the classes that are used by exec, further low any classes they inherit from
// "Level 1": direct usage in async
class ConcatenateInsertCommands extends LineTransform {}
class IgnoreInsecurePasswordWarning extends LineTransform {}
// "Level 2": super classes of Level 1 classes
abstract class LineTransform extends StringTransform {}
// "Level 3" super classes of Level 2 classes
abstract class StringTransform extends Transform {}
```

Depending on how complex the stuff you are doing is, you might want to introduce additional levels, e.g.
```.ts
async function exec() {
  const dump = dumpMysqlDatabase()
  concatInsertCommands(dump)
  ignoreInsecurePasswordWarning(dump)

  function concatInsertCommands(dump) {}
  function ignoreInsecurePasswordWarning(dump) {}
}
```

See also https://github.com/edtr-io/edtr-io/blob/dbd08bad69cdea55c83d4f6fc9d152014e5a0ee8/packages/private/demo/scripts/deploy.ts for another example.

Note: this approach only works for `class` and `function`. Any variable defined with `const` or `let` and therefore any arrow function `const f = () => {}` has to be defined before their first usage. But you can at least move them in a deeper scope resp. abstract them away in functions where it makes sense.

inyono: @kulla FYI, some suggestion on how to improve code readability

Issue #123, opened 2019-12-01
fonts load from de.serlo.localhost:4567 in dev


inyono:
Does only occur in dev (since ASSET_PREFIX == NEXT_ASSET_PREFIX in production): the font isn't loaded correctly, since it is requested via NEXT_ASSET_PREFIX instead of ASSET_PREFIX

Entkenntnis:
fonts get loaded multiple times, two times correctly and two times unprefix - no idea why this is happening

Entkenntnis:
found out the reason, we need to pass assetPrefix to the client

Issue #126, opened 2019-12-01
Implement health check


inyono:
Kubernetes requires services to implement health checks to enable all its functions. Please implement a health check for the frontend service similar to https://github.com/serlo/serlo.org/blob/master/packages/public/editor-renderer/src/index.ts#L45

Entkenntnis:
`/` => `frontend@0.1.1`

inyono:
Please don't close issues before the commits that resolve them are merged into master. You can automate this by including "Close #126" in your commit message. Then GitHub will automatically close the issue after merge.
Also: please open a PR for everything that is done / deployable.

Issue #125, opened 2019-12-01
Review unnecessary files


inyono:
Is https://github.com/serlo/serlo.org/blob/master/packages/public/frontend/server.js still needed?

Entkenntnis:
nope

inyono:
see https://github.com/serlo/serlo.org/issues/126#issuecomment-561103069

Issue #124, opened 2019-12-01
Don't resolve static assets in production


inyono:
Currently, `serverless.js` does resolve static assets when the request isn't handled by a page. While this is currently needed in dev (to serve .next/static), we shouldn't do this in production. For example, you could define a special serverless.js for dev (that basically extends the serverless.js for prod) and add a different entry point here: https://github.com/serlo/serlo.org/blob/master/packages/public/frontend/Dockerfile#L8. Feel free to suggest better solutions

Entkenntnis:
added env-var `SERVE_ASSETS` which is only set to true in development

inyono:
see https://github.com/serlo/serlo.org/issues/126#issuecomment-561103069

PR #132, opened 2019-12-04
Hotfixes for frontend

Entkenntnis:
**serlo.org / frontend**

- use static build id
- frontent-assets server in dev sets cors headers
- comments-page excluded
- set assetPrefix in frontend client
- refactor serverless.js

Closes: #123 #126 #125 #124


packages/public/frontend/docker/FrontendAssetsDockerfile
```diff
@@ -0,0 +1,5 @@
+FROM nginx
+
+COPY packages/public/frontend/docker/nginx.conf /etc/nginx/conf.d/default.conf
+
+ENTRYPOINT ["nginx", "-g", "daemon off;"]
```
inyono: Why do we need this Dockerfile?

packages/public/frontend/next.config.js
```diff
@@ -8,10 +8,8 @@ const withBundleAnalyzer = require('@next/bundle-analyzer')({
 
 module.exports = withCSS({
   target: 'serverless',
-  generateBuildId: async () => {
-    // For example get the latest git commit hash here
-    return require('./package.json').version
-  },
+  generateBuildId: async () =>
+    process.env.BUILD_ID ? process.env.BUILD_ID : null,
```
inyono: Why do you prefer that over the version?

packages/public/frontend/serverless.js
```diff
@@ -16,47 +38,38 @@ const map = {
...
-    return footer.render(req, res)
-  }
-  if (pathname === '/__comments') {
-    return comments.render(req, res)
-  }
+// ONLY for development
+function serveAssets(pathname, req, res, next) {
+  if (process.env.SERVE_ASSETS === 'true') {
+    // based on the URL path, extract the file extention. e.g. .js, .doc, ...
+    const ext = path.parse(pathname).ext
```
inyono: If possible, I'd rather have that not part of the production build (as it only is needed for dev). What do you think about an extra file `serverless.development.js` that imports the server from `serverless.js` and adds the additional middleware for routing?

packages/public/frontend/src/provider.component.tsx
```diff
@@ -228,7 +228,7 @@ setDefaultBreakpoints(
 
 // export function getColor<K extends keyof (typeof theme)["global"]["colors"]>( colorName: K): ((typeof theme)["global"]["colors"][K]) {
 
-export const GlobalStyle = createGlobalStyle`
+export const GlobalStyle: any = createGlobalStyle`
```
inyono: Why is this type needed?

packages/public/frontend/next.config.js
Entkenntnis: The asset url already contains the version so it seems to be unnecesary to include it again

inyono: True. But if I understood the documentation correctly, `null` means that the usual non-deterministic nextjs build id is used. So shouldn't we pass something constant instead?

Entkenntnis: The build command in package.json sets it to the static string `current`, is this sufficient?

packages/public/frontend/docker/FrontendAssetsDockerfile
Entkenntnis: I'll try to avoid it

packages/public/frontend/serverless.js
Entkenntnis: This sounds reasonable

packages/public/frontend/next.config.js
inyono: Ah, didn't see that. That's fine. I'd even set the build id here directly since we don't have a reason to change that up anyways.

packages/public/frontend/docker/FrontendAssetsDockerfile
Entkenntnis: none

packages/public/frontend/src/provider.component.tsx
Entkenntnis: for reasons *xd

packages/public/frontend/serverless.js
Entkenntnis: done

packages/public/frontend/docker/Dockerfile
```diff
@@ -1,11 +1,10 @@
...
 WORKDIR packages/public/frontend
-ENV NEXT_TELEMETRY_DISABLED=1
 ENV NEXT_ASSET_PREFIX http://localhost:3000
 ENV ASSET_PREFIX http://localhost:3001
-ENTRYPOINT ["node", "serverless.js"]
+
+ENTRYPOINT ["node", "serverless/serverless.development.js"]
 EXPOSE 3000
 
 FROM development as release
```
inyono: Add
```
 ENTRYPOINT ["node", "serverless/serverless.production.js"]
```

PR #133, opened 2019-12-04
fixing footer styles

FridoSa:
removing conflicts with bootstrap styles


packages/public/frontend/src/footer/about.tsx
```diff
@@ -108,6 +108,12 @@ const TopButton = styled(Button)`
   }
 ` as typeof Button
 
+const LearnMoreButton = styled(Button)`
+  :hover {
+    text-decoration: none;
+  }
+` as typeof Button
```
inyono: I think that you don't need `as typeof Button` since `styled-component` should be able to handle that itself.

PR #154, opened 2019-12-12
server: Fix html type in history.twig

kulla:


packages/public/server/src/module/Ui/templates/entity/repository/history.twig
```diff
@@ -48,7 +48,7 @@
                 </a>)
             </td>
             <td>{{ isCurrent ? '<span class="fa fa-check"></span>' : '' }}</td>
-            <th>{{ revision.get('changes') }}</th>
```
inyono: I think that was intended, since the changes are the title of the row.

kulla:
Okay, I'll have a look at the spec -> I close this issue in the meanwhile since it is not important...

PR #156, opened 2019-12-13
add button to switch to older revisions

Knorrke:
ping #72

### DIFF
--- a/packages/private/edtr-io/package.json
+++ b/packages/private/edtr-io/package.json
@@ -37,6 +37,7 @@
     "@serlo/mathjax": "^0.0.0",
     "axios": "^0.19.0",
     "lerna": "^3.0.0",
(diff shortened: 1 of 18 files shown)


Knorrke:
ping @inyono I think this is in a good shape to be reviewed. Still needs some Ui-work for the placement of the button to switch revisions and the list of revisions. But The logic should all be there :-)

packages/private/edtr-io/package.json
```diff
@@ -37,6 +37,7 @@
     "@serlo/mathjax": "^0.0.0",
     "axios": "^0.19.0",
     "lerna": "^3.0.0",
+    "moment": "^2.0.0",
```
inyono: We should change that to `luxon` (fine by me if we do that in a future PR)

PR #164, opened 2019-12-17
server: Add helper links for reviewing

kulla:
I worked together with @yolo-bird He whiches to have a list of help link directly on the new page `/entity/unrevised`:

2019-12-17-165426_1055x727_scrot

### DIFF
--- a/packages/public/server/src/module/Entity/src/Entity/Controller/EntityController.php
+++ b/packages/public/server/src/module/Entity/src/Entity/Controller/EntityController.php
@@ -25,11 +25,12 @@ namespace Entity\Controller;
 use Entity\Result;
 use Instance\Manager\InstanceManagerAwareTrait;
 use Zend\EventManager\ResponseCollection;
+use Zend\I18n\Translator\TranslatorAwareTrait;
 use Zend\View\Model\ViewModel;
 
 class EntityController extends AbstractController
 {
-    use InstanceManagerAwareTrait;
+    use InstanceManagerAwareTrait, TranslatorAwareTrait;
 
     public function createAction()
     {
@@ -99,22 +100,23 @@ class EntityController extends AbstractController
             $reviewHelpUrl = "https://docs.google.com/document/d/1p03xx2KJrFw8Mui4-xllvSTHcEPi8G1bdC8rGXcH6f8/edit";
         }
 
+        $translator = $this->getTranslator();
         $helpLinks = [
             [
                 "url" => $reviewHelpUrl,
-                "title" => "Guideline for reviewing",
+                "title" => $translator->translate("Guideline for reviewing"),
             ],
             [
                 "url" => "/discussions",
-                "title" => "List of all discussions",
+                "title" => $translator->translate("List of all discussions"),
             ],
             [
                 "url" => "https://community.serlo.org/channel/feedback-requests",
-                "title" => "Channel #feedback-requests in RocketChat",
+                "title" => $translator->translate("Channel #feedback-requests in RocketChat"),
             ],
             [
                 "url" => "https://docs.google.com/forms/d/e/1FAIpQLSfMjWIZZq2_AoHbqNv3AOEjQRBwA8qEZIMJpk5l0vX7w2nwnQ/viewform",
-                "title" => "Questionnaire for reviewers",
+                "title" => $translator->translate("Questionnaire for reviewers"),
             ],
         ];
 
--- a/packages/public/server/src/module/Entity/src/Entity/Controller/EntityController.php
+++ b/packages/public/server/src/module/Entity/src/Entity/Controller/EntityController.php
@@ -65,7 +65,7 @@ class EntityController extends AbstractController
         return true;
     }
 
-    public function unrevisedAction()
+    protected function getUnrevisedRevisionsBySubject()
     {
         $revisions = $this->getEntityManager()->findAllUnrevisedRevisions()->getIterator();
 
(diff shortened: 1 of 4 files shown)


packages/public/server/src/module/Entity/src/Entity/Controller/EntityController.php
```diff
@@ -86,7 +86,41 @@ public function unrevisedAction()
             }
         }
 
-        $view = new ViewModel(['revisionsBySubject' => $revisionsBySubject]);
+        return $revisionsBySubject;
+    }
+
+    protected function getReviewHelpLinks()
```
kulla: I have added two new functions so that `unrevisedAcion()` does not get to big. However it is also possible to merge those functions. I don't know, what the best solution is...

packages/public/server/src/module/Entity/src/Entity/Controller/EntityController.php
```diff
@@ -86,7 +86,41 @@ public function unrevisedAction()
...
+    {
+        $instance = $this->getInstanceManager()->getInstanceFromRequest();
+
+        if ($instance->getLanguage()->getCode() == "de") {
+            $reviewHelpUrl = "/140479";
+        } else {
+            $reviewHelpUrl = "https://docs.google.com/document/d/1p03xx2KJrFw8Mui4-xllvSTHcEPi8G1bdC8rGXcH6f8/edit";
```
inyono: IMHO: should be on `serlo.org` in the long-term (@yolo-bird) since it might seem kinda sketchy if we link to Google Docs. This is fine for internal stuff, official community documents should be on the platform itself.

packages/public/server/src/module/Entity/src/Entity/Controller/EntityController.php
```diff
@@ -86,7 +86,41 @@ public function unrevisedAction()
...
+            [ "/discussions", "List of all discussions"],
+            [
+                "https://community.serlo.org/channel/feedback-requests",
+                "Channel #feedback-requests in RocketChat",
+            ],
+            [
+                "https://docs.google.com/forms/d/e/1FAIpQLSfMjWIZZq2_AoHbqNv3AOEjQRBwA8qEZIMJpk5l0vX7w2nwnQ/viewform",
+                "Questionnaire for reviewers",
+            ],
+        ];
```
inyono: It feels like that should be translatable / instance-aware. RocketChat is mostly de currently and the feedback form is also in German. Also: possibly hide stuff depending on role. Similarly to above: it feels kinda sketchy that we redirect community reviewers to Google Forms.

packages/public/server/src/module/Ui/templates/entity/unrevised.twig
```diff
@@ -21,6 +21,13 @@
  #}
 <div class="editor-design">
 {{ pageHeader('Unrevised learning resources').render() }}
+<h2>{% trans %} Support links for reviewing {% endtrans %}</h2>
+<ul>
+  {% for helpLink in helpLinks %}
+    {% set url, title = helpLink[0], helpLink[1] %}
```
inyono: Probably easier to read as an object (helpLink['title'], helpLink['url'])

packages/public/server/src/module/Ui/templates/entity/unrevised.twig
```diff
@@ -21,6 +21,13 @@
...
+    <li><a href="{{ url }}">{{ title | trans }}</a></li>
```
inyono: That won't work since title is dynamic

packages/public/server/src/module/Entity/src/Entity/Controller/EntityController.php
kulla: RocketChat will be also used by non-german communities (see for example https://community.serlo.org/channel/tamil ).

@yolo-bird : Do you have on your todo list to translate the questionnaire or shall we include this link only in the German Serlo version? Do you want to hide links for some roles?

packages/public/server/src/module/Entity/src/Entity/Controller/EntityController.php
kulla: It's planned to do an version on en.serlo.org @yolo-bird , right?

packages/public/server/src/module/Ui/templates/entity/unrevised.twig
```diff
@@ -21,6 +21,12 @@
  #}
 <div class="editor-design">
 {{ pageHeader('Unrevised learning resources').render() }}
+<h2>{% trans %} Support links for reviewing {% endtrans %}</h2>
+<ul>
+  {% for helpLink in helpLinks %}
+    <li><a href="{{ helpLink['url'] }}">{{ helpLink['title'] | trans }}</a></li>
```
inyono: That won't work either since `helpLink['title']` is still dynamic

kulla: Fixed with https://github.com/serlo/serlo.org/pull/164/commits/f38421482a664e3b94ce052ddc348c61817a0ee4

packages/public/server/src/module/Ui/templates/entity/unrevised.twig
kulla: Fixed with https://github.com/serlo/serlo.org/pull/164/commits/f38421482a664e3b94ce052ddc348c61817a0ee4

packages/public/server/src/module/Entity/src/Entity/Controller/EntityController.php
yolo-bird: > IMHO: should be on `serlo.org` in the long-term (@yolo-bird) since it might seem kinda sketchy if we link to Google Docs. This is fine for internal stuff, official community documents should be on the platform itself.

The google doc will only be active until the end of february. It's is a nice and easy way to get feedback for the most urgent wishes of reviewers. If @inyono  does not feel great about linking to google docs at all, we'll have to think of another solution.

@kulla  we will add english "subtitles" to the german form, so that at the end it will be bilingual. Hence it will be the same link for every language version. The "subtitles" will be added until this friday.

inyono: Fine by me if it's only temporary

inyono:
Gonna update the i18n files in #194

PR #174, opened 2020-01-02
generate id per component instance

Entkenntnis:


Entkenntnis:
Fixes #176

packages/public/frontend/pages/_document.js
```diff
@@ -146,7 +147,7 @@ class MyNextScript extends NextScript {
       return (
         <>
           <script
-            id={'__NEXT_DATA__' + page}
+            id={'__NEXT_DATA__' + page + '#' + props.pageProps.componentID}
```
inyono: Do we need the page in the id?

Entkenntnis: yes, for now we need it because the client needs to know which page it has to load and this is determined by the name of the root.

PR #175, opened 2020-01-02
fixing some errors in footer component

Entkenntnis:


Entkenntnis:
local build is working, why is this test failing?

packages/public/frontend/package.json
```diff
@@ -12,6 +12,7 @@
     "lint": "yarn tsc --noemit",
     "serverless": "cross-env NEXT_ASSET_PREFIX=http://localhost:3000 ASSET_PREFIX=http://localhost:3000 node serverless/serverless.development.js",
     "start": "cross-env NEXT_ASSET_PREFIX=http://localhost:3000 ASSET_PREFIX=http://localhost:3000 next",
+    "start-dev": "cross-env NEXT_ASSET_PREFIX=http://localhost:3000 ASSET_PREFIX=http://localhost:3001 next",
```
inyono: Why do we need to commands (`start` and `start-dev`) here?

packages/public/frontend/src/assets.ts
```diff
@@ -1,4 +1,8 @@
-const assetPrefix = process.env.ASSET_PREFIX || ''
+let assetPrefix = process.env.ASSET_PREFIX || ''
+
+export function setAssetPrefix(path) {
```
inyono: Why is this needed? Shouldn't this always be `process.env.ASSET_PREFIX`?

packages/public/frontend/package.json
Entkenntnis: The start-dev command uses the frontend-assets service in docker setup, so its useful for me, maybe change the name to something like start-with-asset-server ...

packages/public/frontend/src/assets.ts
Entkenntnis: Unfortunately not when this code is run on the client, so we need to set it explicitly

packages/public/frontend/package.json
inyono: Mh... But isn't `next` already serving the assets in development anyways?

packages/public/frontend/src/assets.ts
inyono: I see. Then why is this different compared to the other usages of the asset prefix? So far, we pass those via props, don't we?

Furthermore: it seems like there are already a couple usages of the assetPrefix in the project. What do you think about an `assetPrefixContext` that we pretty much set once at the root? And every component that needs it (doesn't matter how deep in the tree), uses that context.

Entkenntnis: I haven't decided yet which way to go, a context would be great and maybe it can we implemented at nextjs level, so pages don't need to care about it ... I feel this would fit into a separate PR

inyono: That's fine by me. But the current usages should be done in a similar way, right?

packages/public/frontend/package.json
Entkenntnis: sadly this leads to CORS issues

packages/public/frontend/src/assets.ts
Entkenntnis: done

PR #177, opened 2020-01-03
avoid react warning about html mismatch

Entkenntnis:


packages/public/frontend/pages/__editor.tsx
```diff
@@ -20,6 +20,14 @@ function getCsrfToken(): string {
 
 export default function Index(props) {
   if (typeof window === 'undefined') return null
+
+  const [loading, setLoading] = React.useState(true)
```
inyono: React hooks need to be on the top level and have to occur in all branches (e.g. not after an `if`)

PR #179, opened 2020-01-04
server: add Redis as key-value store

inyono:


docker-compose.yml
```diff
@@ -25,10 +28,12 @@ services:
       - OIDC_SUBJECT_IDENTIFIERS_ENABLED=public,pairwise
       - OIDC_SUBJECT_IDENTIFIERS_PAIRWISE_SALT=youReallyNeedToChangeThis
 
+  #  editor-renderer:
```
inyono: Deactivated `editor-renderer` and `legacy-editor-renderer` local builds for now since they aren't touched in nearly all cases anyways and will be replaced soon by the frontend. And if you want to change them, you want to start them locally anyways (via `yarn start`). Still: previous code is still here if you need it.

packages/public/server/src/module/Frontend/src/Frontend/Factory/RenderComponentHelperFactory.php
```diff
@@ -22,22 +22,19 @@
...
-     * @return mixed
-     */
     public function createService(ServiceLocatorInterface $serviceLocator)
     {
+        /* @var ServiceLocatorInterface $serviceLocator */
         $serviceLocator = $serviceLocator->getServiceLocator();
-        $url   = $serviceLocator->get('config')['services']['frontend'];
-        return new RenderComponentHelper($url);
+        /* @var RenderComponentService $service */
+        $service = $serviceLocator->get(RenderComponentService::class);
```
inyono: Extracted the logic into its own service since `$serviceLocator` behave differently in view helpers vs. service manager. So we should only use view helpers in the view and use the service in controllers / other services / etc. (cc @Knorrke since you had a similar issue / question a couple of weeks ago)

packages/public/server/src/module/KeyValueStore/src/KeyValueStore/Factory/KeyValueStoreFactory.php
```diff
@@ -0,0 +1,60 @@
...
+use Zend\ServiceManager\ServiceLocatorInterface;
+
+class KeyValueStoreFactory implements FactoryInterface
+{
+    public function createService(ServiceLocatorInterface $serviceLocator)
+    {
+        $config = $serviceLocator->get('Config')['key_value_store'];
+        /* @var Service $featureFlags */
+        $featureFlags = $serviceLocator->get(Service::class);
+        if ($featureFlags->isEnabled('key-value-store')) {
```
inyono: hidden behind feature flag

packages/public/server/src/module/Renderer/src/Renderer/Renderer.php
```diff
@@ -99,20 +99,24 @@ public function __construct(FeatureFlagsService $featureFlags, $editorRendererUr
...
+        $key = 'renderer/' . hash('sha512', $input);
+
         if ($this->featureFlags->isEnabled('frontend-legacy-content') && $this->getFormatHelper()->isLegacyFormat($input)) {
-            return $this->renderComponentHelper->__invoke('legacy-content', [
-                'input' => $input,
-            ]);
+            return $this->renderComponentService->render(
+                'legacy-content',
+                ['input' => $input],
+                $key
```
inyono: Unfortunately, this cache handling is still not perfect since we don't know the revision / revision field that is rendered here so we can't have an intelligent cache key. In the future, we probably have a React Component in the frontend that handles the rendering of a whole entity and caches the stuff together.

PR #182, opened 2020-01-04
Use Edtr.io by default

### LINKED ISSUE #159: Activate new Editor for all edits
- For all entities and user profile
- fallback to old editor, if revision was created with the old editor.

### LINKED ISSUE #222: e2e: add tests for every content type and important page
Add tests for the following combinations:
- All content types + taxonomy term + user profile
- created with legacy editor + created with editor
- rendering + editing
- legacy stack + frontend

Knorrke:
This PR includes [edtr.io](https://edtr.io) by default for entities, taxonomy, user and new pages.

Also closes https://github.com/serlo/athene2/issues/882

Fixes #159 Fixes #222


inyono:
What's the state of this PR @Knorrke? CI is failing

packages/public/server/__tests-e2e__/_utils.ts
```diff
@@ -20,23 +20,288 @@
...
+}
+export async function organizeTaxonomy(topic: ElementHandle) {
+  return await getByText(topic, 'Organize taxonomy').then(clickForNewPage)
+}
+
+export async function saveRevision(createPage: ElementHandle) {
+  await getBySelector(createPage, '#subject-nav-wrapper .fa-save').then(click)
+  await getByLabelText(createPage, 'Änderungen').then(e => e.type(randomText()))
+  await createPage.$$('input[type=checkbox]').then(c => c[0].click())
+  await createPage.$$('input[type=checkbox]').then(c => c[3].click())
```
inyono: Should add labels for that and query via label

packages/public/server/__tests-e2e__/_utils.ts
```diff
@@ -20,23 +20,288 @@
...
+
+  return await getByText(createPage, 'Speichern', {
+    selector: 'button'
```
inyono: Why do we need a selector? here

packages/public/server/__tests-e2e__/_utils.ts
```diff
@@ -20,23 +20,288 @@
...
+  }).then(clickForNewPage)
+}
+
+export function randomText(prefix?: string): string {
```
inyono: ?

packages/public/server/__tests-e2e__/_utils.ts
```diff
@@ -20,23 +20,288 @@
...
+  return (prefix ?? 'Text') + ' ' + String(Math.floor(Math.random() * 1e12))
+}
+
+export async function toHaveAttribute(
```
inyono: Instead of export all these, we should have a file `_extend-jest.js` or something like that and call `expect.extend` there. Because now you split the declaration and types into `jest.setup-e2e.ts` and `_utils.ts` (so you need to touch multiple places if you change the signature / add new matchers)

packages/public/server/__tests-e2e__/entities/applet.ts
```diff
@@ -0,0 +1,102 @@
...
+  await expect(applet).toMatchElement('*', { text: appletDescription })
+  await expect(applet).toHaveTitle(`${appletTitle} (applet)`)
+
+  const iframe = await getBySelector(applet, 'iframe')
+  await expect(iframe).toHaveAttribute('src', geogebraUrl)
+})
+
+describe('create/update applet pages', () => {
+  afterEach(async () => {
+    await logout()
```
inyono: If we do it like that, we should also have an `beforeEach` that does the login

packages/public/server/__tests-e2e__/entities/applet.ts
```diff
@@ -0,0 +1,102 @@
...
+    const appletUrlField = await getBySelector(
+      createPage,
+      '#editor article > div:nth-child(1) > div > div'
+    )
+    await appletUrlField.click()
+    await appletUrlField.type(`https://www.geogebra.org/m/${appletId}`)
+
+    const descriptionField = await getBySelector(
+      createPage,
+      '#editor article > div:nth-child(2) > div > div'
```
inyono: Might think about adding some data attributes to Edtr.io so that we can easily query documents

packages/public/server/__tests-e2e__/entities/article.ts
```diff
@@ -0,0 +1,126 @@
...
+  openDropdownMenu
+} from '../_utils'
+import { pages, notifications, elements } from '../_config'
+
+const articleItemType = 'http://schema.org/Article'
+
+test('view article', async () => {
+  const path = '/math/example-content/example-topic-1/example-article'
+  const title = 'Example article'
+  const content = 'Hello World! 42'
```
inyono: The state of that article is wrong

packages/public/server/__tests-e2e__/entities/article.ts
```diff
@@ -0,0 +1,126 @@
...
+  const articlePage = await goto(path)
+  await expect(articlePage).toHaveTitle(title)
+
+  const article = await getByItemType(articlePage, articleItemType)
+  await expect(article).toMatchElement('h1', { text: title })
+
+  const description = await getByItemProp(article, 'articleBody').then(div =>
+    getBySelector(div, 'p')
+  )
+  await expect(description).toHaveHTMLContent(content)
```
inyono: Not a big fan of that since it's heavily dependent on the DOM structure. IMHO, we shouldn't care at what level the description is

packages/public/server/__tests-e2e__/entities/article.ts
inyono: Might be because of the breaking changes in the plugin states we deployed recently. So you probably need to update your example content types in the dump

packages/public/server/docker-entrypoint-initdb.d/001-init.sql
```diff
@@ -1,8 +1,8 @@
--- MySQL dump 10.13  Distrib 5.7.28, for Linux (x86_64)
+-- MySQL dump 10.13  Distrib 5.7.26, for Linux (x86_64)
 --
 -- Host: localhost    Database: serlo
 -- ------------------------------------------------------
--- Server version	5.7.28
+-- Server version	5.7.26
```
inyono: We shoud probably split the fixtures intended for testing from the dump for a couple of reasons:
- We can't update the anonymized dump easily
- We can't run the e2e tests in staging (since they don't have that stuff)
- It's hard to update the test data

There are a couple of things we could do:
- Add an instance for testing purposes via migrations (this could even be in staging & production, too)
- (probably preferable): add another database (with minimal schema / data) that is intended for automated testing, generated via migrations. This way, the anonymized dump can be exchanged / kept up to date. And the testing database can be rolled back whenever (maybe even similar to django after every test), should also be easily applicable to staging / production deployments

packages/public/server/__tests-e2e__/entities/applet.ts
Knorrke: We want to login with different accounts, so moving this to `beforeEach` doesn't work... We thought about making logout more intelligent (e.g. logout only if logged in), and running it after every test.

packages/public/server/__tests-e2e__/entities/applet.ts
Knorrke: yes please. Follow up imho.

packages/public/server/__tests-e2e__/_utils.ts
Knorrke: Probably because the text "Speichern" is displayed in some paragraph as well. But I'll check

jest.config-e2e.js
```diff
@@ -19,13 +19,14 @@
...
  */
-const { mergeDeepRight } = require('ramda')
-
 const puppeteerPreset = require('jest-puppeteer/jest-preset')
 const tsPreset = require('ts-jest/jest-preset')
 
 module.exports = {
-  ...mergeDeepRight(puppeteerPreset, tsPreset),
+  ...puppeteerPreset,
+  ...tsPreset,
```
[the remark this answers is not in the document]
Knorrke: ping @kulla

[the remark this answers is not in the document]
kulla: Yes, this was intentional. In the previous code the setting `setupFilesAfterEnv` of `puppeteerPreset` was overwritten and thus we couldn't use https://github.com/smooth-code/jest-puppeteer/blob/master/packages/expect-puppeteer/README.md#api

`mergeDeepRight` does not concat list properties but uses the list of the second dictionary. Therefore I have deleted this function so that developers are aware that they have to look at the dictionaries and how the configs needed to be merged. (In the case of `puppeteerPreset` and `tsPreset` they have different properties).

A better solution and a Follow-Up is a mergeConfig function based on `R.mergeWith` which concats list and can be used here.

packages/public/server/__tests-e2e__/_config.ts
```diff
@@ -20,3 +23,65 @@
...
  */
 export const exampleApiParameters = ['contentOnly', 'hideBanner', 'fullWidth']
+
+export const elements = {
+  getDropdownToggle(page: ElementHandle) {
+    return getBySelector(page, '#subject-nav-wrapper button.dropdown-toggle')
+  },
+  getLoginButton(page: ElementHandle) {
+    return getByText(page, 'Log in', {
+      selector: '#serlo-menu a'
```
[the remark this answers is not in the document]
kulla: There are two menus currently (one for desktop mode and one for mobile view with the mobile view menu set to `display:none`). Without `#serlo menu` the function `getByText` complains that it founds more than one element. -> Should not be the case with the new redesign

packages/public/server/__tests-e2e__/entities/article.ts
kulla: You mean that only testing whether a text is present is better?

jest.config-e2e.js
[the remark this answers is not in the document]
kulla: Okay, I have added a `mergeConfig` doing what is needed.

inyono:
Merged & deployed the fix for the semantic solution plugin last night. So this PR should be done after we updated the fixtures to the new plugin states.

kulla:
The e2e-tests are passing again (please check it on your computer since i sometimes had problems in the headless mode (in non-headless mode everything worked); I guess due to not waiting enough for DOM changes to take place). From my point of view the most important change requests are implemented and this PR can be merged (Follow ups can be collected in a new issue).

packages/public/server/docker-entrypoint-initdb.d/001-init.sql
inyono: Follow up issue https://github.com/serlo/serlo.org/issues/293

packages/public/server/__tests-e2e__/entities/article.ts
inyono: yes

inyono:
Follow ups: #293 #294 #295

PR #185, opened 2020-01-06
Some english phrases, that are part of the user-output are translated in a weird way into german

jakobwes:
For example "Groups" is translated by "Aufgaben", so means in fact "Exercises". This leads to the point that communities in new language-version who translate the platform starting from the english-crowdin translate then "Gruppen" literally. I changed the english user-output, so that if the new phrases (the phrases that I added/changed) are pushed to crowdin the english-output and therefore also the output in all other language-version (translated from the english-output) is more correct.

Following output-phrases have been changed (behind "-->" is the new term that needs to be added to crowdin with the respective german translation)

Groups - Aufgaben --> Exercises - Aufgaben
Update a page - Seite bearbeiten --> Update page - Seite bearbeiten
Edit About Horizon - "Über Horizont" bearbeiten --> Edit "About Horizon" - "Über Horizon" bearbeiten
Discuss - Kommentieren --> Comment - Kommentieren
Manage related links - Zugehörige Inhalte verwalten --> Manage related Content - Zugehörige Inhalte verwalten
Organize - Baumstruktur bearbeiten --> Organize taxonomy - Baumstruktur bearbeiten
unpublished - Unveröffentlichte Blogeinträge --> unpublished blog-posts - Unveröffentliche Blogeinträge

I only changed output-phrases where the literal meaning is different in english and german. 
**Just for completion:** these phrases in Crowdin also have other meanings in the german translation, but I didn't find them (at least in templates/UI)
You have downvoted this comment. - Du hast diesen Kommentar bereits abgewertet. --> nicht gefunden
You have upvoted this comment. - Du hast diesen Kommentar bereits aufgewertet. --> nicht gefunden
applet - GeoGebra Applet --> nicht gefunden

And there the meaning is slightly different (nice to have), but it isn't preasuring: 
Flags - Gemeldete Inhalte
The requested controller was not dispatchable. - Der angeforderte Controller konnte nicht abgesendet werden.
input-string-normalized-match-challenge - Eingabeaufgabe: Wort
multiple-choice-right-answer - richtige Multiple Choice-Antwort
multiple-choice-wrong-answer - falsche Multiple Choice-Antwort
Unarchive discussion - Diskussion wiederherstellen
Current password - Aktuelles Passwort eingeben
New password - Neues Passwort eingeben
Reply - Antwort Abschicken
Password - Passwort eingeben:
Submit - Kommentar abschicken
This site makes heavy use of JavaScript! - Diese Seite benötigt JavaScript!
Type answer - Deine Lösung
[...]

PR #194, opened 2020-01-10
i18n improvements

inyono:
Supersedes #185 because I can't force-push to https://github.com/jakobwes/serlo.org


inyono:
@jakobwes For the future: you can check the comments in https://github.com/serlo/serlo.org/blob/master/packages/public/server/src/lang/athene2.po to see all the places where a string is used.

inyono:
Translation files have been updated on CrowdIn. Please translate the new strings so that they are part of the next deployment (cc @yolo-bird @jakobwes)

PR #195, opened 2020-01-10
make automatic checkout for reviewers explicit and opt-in

Knorrke:


packages/public/client/src/main/index.js
```diff
@@ -199,7 +199,8 @@ const init = $context => {
       initEntityEditor(
         {
           initialState: $editor.data('state'),
-          type: $editor.data('type')
+          type: $editor.data('type'),
+          mayCheckout: $editor.data('may-checkout')
```
inyono: Is this really a boolean?

inyono:
Btw: Thanks for changing it in the frontend package, too

packages/public/client/src/main/index.js
Knorrke: well.... it's a 1 or 0 atm? :sweat_smile:

PR #198, opened 2020-01-10
fixing incorrect minifier

Entkenntnis:
multi-pass of terser is breaking existing code


Entkenntnis:
maybe reseting redis cache is necessary

PR #227, opened 2020-01-15
server: Fix validation of applet urls (fixes #196)

### LINKED ISSUE #196: edtr-io: saving of applets not working correctly
```
url: {notUri: "The input does not appear to be a valid Uri",…} notUri: "The input does not appear to be a valid Uri" regexNotMatch: "Applet-URL invalid. Use one of the form geogebra.org/m/id or ggbm.at/id"
```

kulla:
All occurrences of ggbm.at are replaced at serlo (there is only one version left which is trashed).


Knorrke:
@inyono done

Issue #166, opened 2019-12-21
Inspect strange encoding in 001-init.sql


kulla:
Today (okay, yestderday :-) ) while working on #129 I noticed that my dumps messed with the encodings and change them between dumps of `001-init.sql`. I had a look in the current [`001-init.sql`](https://github.com/serlo/serlo.org/blob/master/packages/public/server/docker-entrypoint-initdb.d/001-init.sql) and noticed that `ö` is encoded as `C3 83 C2 B6`:

![2019-12-21-044144_931x196_scrot](https://user-images.githubusercontent.com/1327215/71302535-3ba15f80-23ac-11ea-9333-9b330320e038.png)

However the right [utf8 encoding of `ö`](https://en.wikipedia.org/wiki/%C3%96#Encoding) is `C3 B6`. Concerning [this post](https://www.postgresql.org/message-id/flat/D960CB61B694CF459DCFB4B0128514C207BB6119%40exadv11.host.magwien.gv.at#8bca8aca86a737abecf96f329614bea6) this happens when the Database using `utf8` works with a client using `Latin1`. The Database sends `C3 B6` which is `Ã¶` in `Latin1`. When this is resend to the server it gets encoded to `C3 83 C2 B6` which is `Ã¶` in `utf8`. This explains why I see `kÃ¶nnen` when I have a look at `001-init.sql` with `less`:

![2019-12-21-044806_305x95_scrot](https://user-images.githubusercontent.com/1327215/71302586-182ae480-23ad-11ea-8e43-790cc3cdc04e.png)

Unfortunately already [the first commit of this file}(https://github.com/serlo/athene2/blob/09be8b8cc747bef5a62ad411a0f60c45d79f908f/docker-entrypoint-initdb.d/001-init.sql) in the package `athene2` contains this strange encoding. I would like to track this issue. When you have an idea, how this happen I am happy when you help me :smile: (This is why I hate dealing with encodings).

kulla:
Okay I guess I have found the answer why we haven't noticed this yet. The default character set for a mysql client seems to be `latin1`. you can check this after `yarn start` via:

````$ docker-compose exec mysql mysql --user=root --password=secret
mysql: [Warning] Using a password on the command line interface can be insecure.
Welcome to the MySQL monitor.  Commands end with ; or \g.
...
mysql> SHOW VARIABLES LIKE '%character_set_results%';
+-----------------------+--------+
| Variable_name         | Value  |
+-----------------------+--------+
[...]

PR #231, opened 2020-01-18
server: change database charset to utf8mb4

inyono:
PoC for #166. `yarn migration:up` and `yarn migration down` works correctly for our local dump. Haven't tested it with a dump from production, yet.


inyono:
Welp, still need to handle
> InnoDB has a maximum index length of 767 bytes, so for utf8mb3 or utf8mb4 columns, you can index a maximum of 255 or 191 characters, respectively. If you currently have utf8mb3 columns with indexes longer than 191 characters, you will need to index a smaller number of characters

see https://dev.mysql.com/doc/refman/5.5/en/charset-unicode-conversion.html. This concerns e.g. aliases, but maybe other columns, too.

inyono:
see also https://serversforhackers.com/c/mysql-utf8-and-indexing

inyono:
Probably not needed anymore since we want to use the API in the future (instead of working with the DB directly)

PR #273, opened 2020-01-27
Edtr.io 0.50.0

inyono:
- [x] Integrate #187
- [x] Integrate #254
- [x] Migration script
  - [x] script: https://gist.github.com/inyono/399dd88512a11acdcf20f750960110ed
  - [x] Test migration script on production dump

PR #276, opened 2020-01-28
hotfix: use new slate state converter

Knorrke:
Converting is broken due to state change in #273. This fixes the issue, however we should probably refactor our slate converter soon.


Knorrke:
damnit, apparently @inyono beat me to it :D already fixed in https://github.com/serlo/serlo.org/commit/bf2b6f0d506afe96aab2281e5068d4d3730156f9

PR #287, opened 2020-02-03
Recycle bin: filter by language

### LINKED ISSUE #204: i18n: Show in /uuid/recycle-bin only entities of current language version

Lodifice:
Partially fixes #204.
This will filter entities and revisions by language, won't work for other kinds of objects.

TODO: add taxonomies

### DIFF
--- a/packages/public/server/src/module/Uuid/config/module.config.php
+++ b/packages/public/server/src/module/Uuid/config/module.config.php
@@ -124,7 +124,7 @@ return [
                     ],
                     'setInstanceManager' => [
                         'required' => true,
-                    ]
+                    ],
                 ],
             ],
         ],
--- a/packages/public/server/src/module/Uuid/config/module.config.php
+++ b/packages/public/server/src/module/Uuid/config/module.config.php
@@ -122,6 +122,9 @@ return [
                     'setUuidManager' => [
                         'required' => true,
                     ],
+                    'setInstanceManager' => [
+                        'required' => true,
+                    ]
                 ],
             ],
         ],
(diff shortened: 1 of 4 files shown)


packages/public/server/src/module/Uuid/config/module.config.php
```diff
@@ -105,6 +105,9 @@
                     'setUuidManager' => [
                         'required' => true,
                     ],
+                    'setInstanceManager' => [
+                        'required' => true,
+                    ]
```
kulla: @vroland Run `yarn format:php` after you have finished the work on this PR to format the PHP code and to remove the lint error.

inyono:
What's the state of this PR? Ready to review / merge apart from the missing `yarn format`?

Lodifice:
The code for filtering taxonomies by language is still missing. I thought @vroland would add it soon.

vroland:
@inyono Is there a general way to associate some Item (comment, revision, entity, taxonomy item...) with the language version / instance? It doesn't seem very efficient and maintainable to re-implement filtering for every kind of item.

inyono:
Not that i know of. Especially, since (AFAIK) some items aren't directly tied to the instance / don't have a unique instance. E.g. taxonomy can get its instance via its subject, entities via the taxonomies they are linked to, revisions via their entities, etc.

vroland:
Hm, that's unfortunate. I'll look into how to implement it for taxonomy items when I find the time, but I don't think it makes sense right now to handle every possible item.

kulla:
Yes, when we add support for taxonomy it's enough. We can do this tomorrow at our meeting...

PR #322, opened 2020-02-26
moved equations plugin to serlo repo

metzgaria:
not the most beautiful, but it works :woman_shrugging:


inyono:
Closed in favor of #413

PR #333, opened 2020-03-03
Fixing loading font from root path

Entkenntnis:
Ok, it's quite an obvious solution


inyono:
How does this affect SSR?

Entkenntnis:
ssr will work as usual, by "runtime" i mean the react runtime: the font path is loaded when a new globalstyle element is created and not, when globalstyle itself is created (which happens on file-load and at this moment the assetPath is not set yet)

inyono:
I suppose this is no longer relevant because of serlo/frontend, right?

Entkenntnis:
Yes

PR #394, opened 2020-04-24
added topic mathematics to spanish language-version (

jakobwes:
We have an author that works on this topic


packages/public/server/src/module/Subject/config/instances.config.php
```diff
@@ -388,6 +388,21 @@
...
+                        'topic',
+                        'locale',
+                    ],
+                    'allowed_entities'   => [
+                        'article',
+                        'text-exercise',
+                        'video',
+                        'course',
+                        'text-exercise-group',
+                        'math-puzzle',
```
inyono, line 402: We shouldn't use math-puzzle for new stuff.
```suggestion
```

PR #416, opened 2020-05-12
Upgrade jquery to 3.5.1 (Fixes #378)

### LINKED ISSUE #378: Update jQuery to 3.5.0
Follow up of #374: Currently we cannot update jQuery to `3.5.0` since afterwards collapsible wouldn't work any more (see issue https://github.com/twbs/bootstrap/issues/30553 ). We need to wait until https://github.com/twbs/bootstrap/pull/30559 got deployed in a new version (which should be bootstrap `4.4.2`).

kulla:
Fixes #378


inyono:
Already fixed in #413

PR #435, opened 2020-05-28
Add autoreview in predefined taxonomy terms

### LINKED ISSUE #312: Automatic Checkout in the sandbox
New authors should have a place where they can test editing without waiting for their review. 

Edits in de.serlo.org/community/sandkasten and in children taxonomy terms should be checked out automatically. (this issue is only relevant for the german sandbox)

kulla:
Fixes #312 

~*Unhold*: There seems to be a discussion in the issue whether we want to have it.~

### DIFF
--- a/packages/public/server/src/module/Taxonomy/src/Entity/TaxonomyTerm.php
+++ b/packages/public/server/src/module/Taxonomy/src/Entity/TaxonomyTerm.php
@@ -317,7 +317,7 @@ class TaxonomyTerm extends Uuid implements TaxonomyTermInterface
         return $associations->contains($object);
     }
 
-    public function knowsAncestor(TaxonomyTermInterface $ancestor)
+    public function isAncestorOrSelf(TaxonomyTermInterface $ancestor)
     {
         $term = $this;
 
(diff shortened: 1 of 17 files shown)


kulla:
@inyono Ready to be merged :-)

kulla:
@inyono Feedback is implemented. Ready to be merged...

packages/public/server/src/module/Taxonomy/src/Entity/TaxonomyTerm.php
```diff
@@ -320,11 +320,17 @@ public function isAssociated(TaxonomyTermAwareInterface $object)
...
+        while (true) {
             if ($term === $ancestor) {
                 return true;
             }
+
+            if ($term->hasParent()) {
+                $term = $term->getParent();
+            } else {
+                break;
+            }
```
inyono, line 333: Why is this needed? If I understand the change correctly, this now also covers the case where `$this` is already `$ancestor` (instead of any parent). Is this a change that is conform to its usage? If so, should we maybe clarify the name? Not sure if I'd call me my own ancestor 😅

kulla: Yes, this is the new use case -> I have renamed the method...

kulla:
@inyono new feedback included -> ready to be merged

kulla:
@inyono In https://github.com/serlo/serlo.org/pull/435/commits/79d62d0582bbfaec9d4cb5c85cf66f7d1f69f673 I have included a documentation for this feature in `README.md` You can remove this commit when you think this is not necessary. :-)

PR #450, opened 2020-06-22
Quickfix: make viewbox larger for spanish text

### LINKED ISSUE #446: Part of the text missing starting page of es.serlo.org
In es.serlo.org in the graphic:

CopyPasteDesign 2020-06-15 at 19 58 52

a part of the text is missing.

Knorrke:
fixes #446 
Problem is: the viewBox cuts off the spanish text, so I increased the width of the svg box by 10 to left and right.
Maybe there's also a way to automatically set the viewBox, but this is a quick fix for the issue, until a language has even longer phrases :D


inyono:
Yay, SVG Wizard for the rescue. Thanks @Knorrke! 🎉

PR #467, opened 2020-07-31
client: Render GeogebraTube inside an iframe

### LINKED ISSUE #462: GeoGebra integration in legacy content is broken
See https://de.serlo.org/54038. Seems like GeoGebra changed something. We should probably use the same thing we do in Edtr.io, i.e. using the GeoGebra Tube iframe.

So: change legacy editor renderer to use GeoGebra Tube iframe.

kulla:
In legacy markdown content Geogebra is rendered inside an iframe instead of using the Geogebra JS library (fixes #462 ). It mimics the behavior of https://github.com/serlo/frontend/blob/staging/src/components/content/geogebra.tsx 

### Testing:

Go to http://en.serlo.localhost:4567/math/example-content/example-topic-1/example-article-%28old-editor%29 and add `>[Manipulation2](ggt/2931245)` as content. After saving Geogebra is rendered inside an iframe.


kulla:
@inyono Do you know for which kind of GeoGebra applets the function `[initGeogebraApplet](https://github.com/serlo/serlo.org/blob/835418326e43e9d9c88b6f60b3b1ebb285450972/packages/public/client/src/frontend/modules/injections.js#L65)` is used? Is it for the case that we have uploaded a geogebra file to `assets.serlo.org`?!

inyono:
> @inyono Do you know for which kind of GeoGebra applets the function [initGeogebraApplet](https://github.com/serlo/serlo.org/blob/835418326e43e9d9c88b6f60b3b1ebb285450972/packages/public/client/src/frontend/modules/injections.js#L65) is used? Is it for the case that we have uploaded a geogebra file to `assets.serlo.org`?!

Pretty sure that these are legacy XML applets that we hosted ourselves. We shouldn't have those anymore.

PR #497, opened 2020-09-10
Api update on taxonomy order

kulla:
Fixes https://github.com/serlo/api.serlo.org/issues/72

### DIFF
--- a/packages/public/server/src/module/Taxonomy/src/Manager/TaxonomyManager.php
+++ b/packages/public/server/src/module/Taxonomy/src/Manager/TaxonomyManager.php
@@ -377,10 +377,16 @@ class TaxonomyManager implements TaxonomyManagerInterface
         return $term;
     }
 
-    public function triggerUpdate(TaxonomyTermInterface $term)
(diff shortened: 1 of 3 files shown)


packages/public/server/src/module/Taxonomy/src/Manager/TaxonomyManager.php
```diff
@@ -377,6 +377,18 @@ public function updateTerm(FormInterface $form)
...
+    public function flushUpdatesForTerm(TaxonomyTermInterface $term)
+    {
+        $this->flush();
+        // We need to detach $term from the ORM cache and reload a new $term from the database or otherwise the
+        // new children order is not stored
+        $this->objectManager->detach($term);
+        $updatedTerm = $this->getTerm($term->getId());
+        $this->getEventManager()->trigger('update', $this, [
+            'term' => $updatedTerm,
+        ]);
```
kulla, line 389: Alternative: We can move this logic to `updateTerm()` for the parent element which is updated. However then we trigger multiple update events for the parent element.

inyono (approved):

PR #502, opened 2020-09-23
server-api: Add alias support for /user/profile/:username

kulla:
Add support for resolving `/user/profile/:username` in `/api/alias/:alias` so that https://github.com/serlo/api.serlo.org/pull/84 can be finished (see also https://github.com/serlo/api.serlo.org/issues/67 ).

Example result for `http://de.serlo.localhost:4567/api/alias/user/profile/Kulla`:

```json
{
  "id": 26217,
  "instance": "de",
  "path": "/user/profile/Kulla",
  "source": "/user/profile/Kulla",
  "timestamp": "2014-07-07T18:24:52+02:00"
}
```

- [x] return `source: /user/profile/:id`

### DIFF
--- a/packages/public/server/__tests-e2e__/api.ts
+++ b/packages/public/server/__tests-e2e__/api.ts
@@ -1,0 +1,50 @@
+/**
+ * This file is part of Serlo.org.
+ *
+ * Copyright (c) 2013-2020 Serlo Education e.V.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License")
+ * you may not use this file except in compliance with the License
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * @copyright Copyright (c) 2013-2020 Serlo Education e.V.
+ * @license   http://www.apache.org/licenses/LICENSE-2.0 Apache License 2.0
+ * @link      https://github.com/serlo-org/serlo.org for the canonical source repository
+ */
+import axios from 'axios'
+
+import { testingServerUrl } from './_config'
+
+describe('/api/alias/:alias', () => {
+  describe('/api/alias/user/profile/:username', () => {
+    test('when user exists', async () => {
+      const response = await fetchPath('/api/alias/user/profile/admin')
+
+      expect(response.data).toEqual({
+        id: 1,
+        instance: 'en',
+        path: '/user/profile/admin',
+        source: '/user/profile/admin',
+        timestamp: '2014-03-01T20:36:21+01:00',
+      })
+    })
+    
+    test('when user does not exist', async () => {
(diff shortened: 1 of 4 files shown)


inyono (changes requested):
Looks good to me, merging blocked until https://github.com/serlo/api.serlo.org/pull/84 got merged (so we have up-to-date contracts). Depending on what we decide for https://github.com/serlo/api.serlo.org/pull/84#discussion_r499651364, this might need some minor changes.

packages/public/server/src/module/Api/src/ApiManager.php
```diff
@@ -76,6 +77,21 @@ public function getAliasData(AliasInterface $alias)
         ];
     }
 
+    public function getAliasDataFromUser(
```
inyono, line 80:
```suggestion
    public function getAliasDataForUser(
```
reads better IMHO.

inyono: I'd even think about moving the instance part up since it doesn't affect the user (and you just polyfill it so that it works in every instance). So you could do that in the `ApiController` instead (that is aware of instances).

packages/public/server/__tests-e2e__/api.ts
```diff
@@ -0,0 +1,50 @@
...
+      const response = await fetchPath('/api/alias/user/profile/admin')
+
+      expect(response.data).toEqual({
+        id: 1,
+        instance: 'en',
+        path: '/user/profile/admin',
+        source: '/user/profile/1',
+        timestamp: '2014-03-01T20:36:21+01:00',
+      })
+    })
```
kulla, line 38: @inyono This can be deleted after we have the contract tests ( I just used it for refactoring)

packages/public/server/__tests-e2e__/api.ts
```diff
@@ -0,0 +1,50 @@
...
+
+    test('when user does not exist', async () => {
+      const response = await fetchPath('/api/alias/user/profile/not-existing')
+
+      expect(response.data).toBeNull()
+    })
```
kulla, line 44: Note to me for follow up issue: Add a contract test for this case.

packages/public/server/src/module/Api/src/Controller/ApiController.php
```diff
@@ -62,6 +65,28 @@ public function aliasAction()
         $alias = $this->params('alias');
         $instance = $this->getInstanceManager()->getInstanceFromRequest();
 
+        $userProfilePrefix = 'user/profile/';
+        if (strpos($alias, $userProfilePrefix, 0) === 0) {
```
inyono, line 69: `0` seems to be the default for the third argument, so I'd remove it. Furthermore: I'd probably prefer to extract that into a method `startsWith` (or similar) since I wasn't aware of `strpos`.

inyono: Also: a regex might be more readable (especially since you extract the username afterwards anyways). Would make it pretty clear that you are looking for a string 'user/profile/:username'.

inyono (approved):
Thanks, contract tests seem to work. I noted a minor thing (that can also be moved into a follow-up issue). Feel free to merge.

packages/public/server/src/module/Api/src/Controller/ApiController.php
kulla: Thx, I merge it and I'll do a follow up PR...

PR #504, opened 2020-09-28
server-api-uuid: Add field "revisionIds"

kulla:
Supplement of https://github.com/serlo/api.serlo.org/pull/87 for fixing https://github.com/serlo/api.serlo.org/issues/73

### DIFF
--- a/packages/public/server/src/module/Api/src/ApiManager.php
+++ b/packages/public/server/src/module/Api/src/ApiManager.php
@@ -345,8 +345,8 @@ class ApiManager
     {
         return $uuid
             ->getRevisions()
-            ->map(function ($rev) {
-                return $rev->getId();
+            ->map(function ($revision) {
+                return $revision->getId();
             })
             ->toArray();
     }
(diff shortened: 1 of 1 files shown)


packages/public/server/src/module/Api/src/ApiManager.php
```diff
@@ -339,6 +341,16 @@ public function getUuidData(UuidInterface $uuid)
         return $data;
     }
 
+    private function getRevisionIds($uuid)
+    {
+        return $uuid
+            ->getRevisions()
+            ->map(function ($rev) {
+                return $rev->getId();
+            })
```
inyono, line 350: I'd prefer
```suggestion
            ->map(function ($revision) {
                return $revision->getId();
            })
```
I try to avoid as many abbreviations as possible as long as it gets not too verbose.

inyono (changes requested):
Blocked until the PR on api.serlo.org got merged.

kulla:
@inyono This PR is ready to be merged. Together with #513 all contract tests should be fulfilled. Afterwards a deploy is IMHO okay since I need time to implement the events feature...

PR #508, opened 2020-10-07
Add support for /api/events

kulla:
Complementary PR for https://github.com/serlo/api.serlo.org/pull/88


packages/public/server/__tests-e2e__/api.ts
```diff
@@ -0,0 +1,120 @@
...
+
+  test('with ?first=2', async () => {
+    const response = await fetchPath('/api/events?first=2')
+
+    expect(response.data).toEqual({
+      totalCount: 84706,
+      eventIds: [1, 2],
+      pageInfo: { hasNextPage: true, hasPreviousPage: false },
+    })
+  })
```
inyono, line 66: That is not the indented way to implement cursors. Actually, serlo.org doesn't need to be aware of the cursors in any way and just needs to return the whole list every time (api.serlo.org does the filtering etc.). Please take a look at the implementation for the notifications as an example.

inyono (changes requested):

packages/public/server/__tests-e2e__/api.ts
kulla: We have currently ~ 330.000 events. When we do all the filtering at api.serlo.org we need to load (or request) all events for each api call. Thus I wanted the filtering happen at serlo.org. Or do you suggest to do the filtering of `userId` and `uuid` at serlo.org and the rest at serlo.org? Then we would have a big overhead when we have no filters since we would return a list with 330.000 eventIds...

kulla: When you think doing the filtering at api.serlo.org is no big deal, I can implement the filtering there. (I had such a solution already in https://github.com/serlo/api.serlo.org/pull/88/files/2ef840302bec1cea931110b45b5fea46ce27bb5d..493e9babfbd611f11e0c2d1760da5cf98ffa4b54 )

inyono: IMHO:
- Filtering by userId and uuid at serlo.org level makes sense (this is also easily cacheable)
- Outputting all events might be a problem (at least if we handle the caching the same as we did for other resources at the moment). We might want to add a time-based cache to that.
- If we ran into performance problems in the future, we can still add a better cache for specific use cases (e.g. while the list of all events doesn't need up-to-date and can have a time-based cache, we might want an up-to-date list of today's events). I'd initially leave it like we have currently, though (1 request = 1 entry in cache. Smart cache value update where possible, time-based where not).

kulla:
Yeah, we will not implemented this in this repo any more...

PR #526, opened 2020-11-02
Fix spelling

xnumad:
https://de.wiktionary.org/wiki/lizenzieren


inyono:
Thanks :) I noticed that you (or someone else) already made the corresponding changes at Crowdin. The new translation will be part of the next deployment.

PR #531, opened 2020-11-13
server: Fix bug in /api/event/:id

### LINKED ISSUE #519: Error in /api/event/:id
http://de.serlo.localhost:4567/api/event/65 returns the following error:

`Fatal error: Uncaught Error: Call to a member function getId() on null in /usr/local/apache2/htdocs/src/module/Api/src/Manager/NotificationApiManager.php:236 Stack trace: #0 /usr/local/apache2/htdocs/src/module/Api/src/Manager/NotificationApiManager.php(100): Api\Manager\NotificationApiManager->normalizeEvent(Object(Event\Entity\EventLog)) #1 /usr/local/apache2/htdocs/src/module/Api/src/Manager/NotificationApiManager.php(95): Api\Manager\NotificationApiManager->getEventData(Object(Event\Entity\EventLog)) #2 /usr/local/apache2/htdocs/src/module/Api/src/Controller/NotificationApiController.php(57): Api\Manager\NotificationApiManager->getEventDataById(65) #3 /usr/local/apache2/htdocs/src/vendor/zendframework/zendframework/library/Zend/Mvc/Controller/AbstractActionController.php(82): Api\Controller\NotificationApiController->eventAction() #4 [internal function]: Zend\Mvc\Controller\AbstractActionController->onDispatch(Object(Zend\Mvc\MvcEvent)) #5 /usr/local/apache2/htdocs/src/vendor/zendframework/zendframework/library/Zend/E in /usr/local/apache2/htdocs/src/module/Api/src/Manager/NotificationApiManager.php on line 236`

kulla:
Fixes #519 There are events of the kind `taxonomy/term/associate` and
`taxonomy/term/dissociate` where the child uuid might be null (This might
be due to deleted uuids from the database). Also for some events of the
kind `taxonomy/term/parent/change` the previous or the new parent is set
to `"no parent"` in the database. This might be du to a bug or an old
behavior of the Serlo software.

### DIFF
--- a/packages/public/server/src/module/Api/src/Manager/NotificationApiManager.php
+++ b/packages/public/server/src/module/Api/src/Manager/NotificationApiManager.php
@@ -229,18 +229,20 @@ class NotificationApiManager
                     'reason' => $event->getParameter('reason') ?? '',
                 ];
             case 'taxonomy/term/associate':
+                $object = $event->getParameter('object');
                 return [
                     '__typename' => 'CreateTaxonomyLinkNotificationEvent',
                     'actorId' => $event->getActor()->getId(),
                     'parentId' => $event->getObject()->getId(),
-                    'childId' => $event->getParameter('object')->getId(),
+                    'childId' => $object ? $object->getId() : null,
                 ];
             case 'taxonomy/term/dissociate':
+                $object = $event->getParameter('object');
                 return [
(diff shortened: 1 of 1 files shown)


inyono:
Since this is just an database inconsistency that we didn't handle correctly so far, I'd rather fix that in the db directly, see #533.

PR #538, opened 2020-11-17
server: Add "objectId" to /api/event/:id

kulla:
Needed for https://github.com/serlo/api.serlo.org/issues/92

Here is the only exception, where `objectId` is defined differently:

```php
            case 'discussion/create':
                return [
                    '__typename' => 'CreateThreadNotificationEvent',
                    'actorId' => $event->getActor()->getId(),
                    'threadId' => $event->getObject()->getId(),
                    'objectId' => $event->getParameter('on')->getId(),
                ];
```


inyono:
Making this into a draft PR until the contracts are live.

inyono (approved):
Thanks, Contract tests succeed.

PR #540, opened 2020-11-24
Rework alias system

inyono:
This PR overhauls and simplifies our alias system.

### Changed

- No longer redirect from `/:old-alias` or `/:id` to `/:current-alias`. I.e.: the legacy system resolves all aliases pointing to an object the same way. The Cloudflare worker is responsible for making sure that we have the appropriate redirects to avoid duplicate content (which is already in-place).
- Dynamically create aliases for objects. Basically all aliases of the form `/optional-prefix/id/readable-title` work. The `optional-prefix` is usually the subject where we can determine it (edge cases: pages, see also discussion below), `readable-title` is dependent on the entity itself.
- In our slugify logic, we now remove a couple of additional characters that have special meaning in URLs just to make sure that things work even when paths aren't encoded.

### Removed

- Removed the possibility to set the URL of static pages in the backend (since those aliases are now automatically generated the same way as for other objects).

### Breaking Changes

- We no longer return `timestamp` and `source` via alias API. If quickly glanced at the implementation in Frontend and Cloudflare Worker and I'm pretty sure we don't use those fields at the moment.

## Open Questions

- When we remove the legacy aliases in the future, we break use cases where we rely on the actual alias. This should only be the case for static pages since those where the only cases where one could actually determine the alias (i.e. /serlo, /mathe, ...).   Basically /serlo resp. /mathe move to /:id/serlo resp. /:id/mathe-startseite
-> Since those cases are quite few anyways, I'd say we could hard-code them in the Cloudflare Worker. Most likely, we'll remove most static pages anyways (to allow custom rendering) and for most other stuff the actual alias isn't really important anyways.


inyono:
Merging this so I can deploy this in staging. Please still post any feedback, I'll address them before a potential production release.

packages/public/server/src/module/Alias/src/Controller/AliasController.php
```diff
@@ -22,71 +22,149 @@
...
+            $url = $this->resolveAlias($alias);
+            return $url === null
+                ? $this->notFoundResponse()
+                : $this->routerResponse($url);
         }
+    }
 
+    private function resolveAlias($alias)
+    {
+        if (preg_match('/(?<id>\d+)\//', $alias, $matches)) {
```
kulla, line 81: This is a problem, when there is a number in the subject's name (e.g "Analysis 1"). I have renamed "Physik" to "Physik 1" -> now https://de.serlo-staging.dev/physik/ resolves to aeneas profile...

kulla: A first workaround:

```suggestion
        if (preg_match('/\/(?<id>\d+)\//', "/" . $alias, $matches)) {
```

However we still have an issue when the subject's name consists only of digits

packages/public/server/src/module/Alias/src/Controller/AliasController.php
```diff
@@ -22,71 +22,149 @@
...
-            ])
-        );
+        return $this->forward()->dispatch($controller, $params);
+    }
 
-        return $return;
+    private function notFoundResponse()
+    {
+        $this->getResponse()->setStatusCode(404);
+        return false;
```
kulla, line 168: Follow up: I would move this function to an helper Mixin / base class so that we can use it in other controllers as well.

packages/public/server/src/module/Common/src/Filter/Slugify.php
```diff
@@ -28,34 +28,51 @@ class Slugify implements FilterInterface
 {
     /**
      * @param string $text
-     * @return bool|mixed
+     * @return bool|string
      */
-    protected static function slugify($text)
+    protected static function slugify(string $text)
```
kulla, line 33: Why is this function prototected? I would do one of the followings:

1. Make this function public and call it in AliasManager via `Slugify::slugify(...)` > maybe move the static function to `Common\Utils` where we can collect utility functions
2. Merge this function into the `filter()` function

I prefer (1) as a solution

packages/public/server/src/module/Common/src/Filter/Slugify.php
```diff
@@ -28,34 +28,51 @@ class Slugify implements FilterInterface
...
     {
-        $text = trim($text, ' ');
```
kulla, line 35: Don't we want to have a trim of the text for articles with white spaces at the beginning or end of the name?

packages/public/server/src/module/Common/src/Filter/Slugify.php
```diff
@@ -28,34 +28,51 @@ class Slugify implements FilterInterface
...
-     */
-    protected static function toAscii($str, $delimiter = '-')
-    {
-        $clean = iconv('UTF-8', 'ASCII//TRANSLIT', $str);
-        $clean = preg_replace('/[^a-zA-Z0-9\/_|+ -]/', '', $clean);
-        $clean = strtolower(trim($clean, '-'));
-        $clean = preg_replace('/[\/_|+ -]+/', $delimiter, $clean);
-
-        return $clean;
-    }
```
kulla, line 57: Sidenote: I want to have a dead code detector so that we can get rid of all the dead code in this repo. Thanks for deleting :tada: :smile:

packages/public/server/src/module/CacheInvalidator/config/module.config.php
```diff
@@ -42,49 +42,4 @@
...
-    //                ],
-    //                'parameter.update' => [
-    //                    __NAMESPACE__ . '\Invalidator\NavigationStorageInvalidator',
-    //                ],
-    //                'parameter.remove' => [
-    //                    __NAMESPACE__ . '\Invalidator\NavigationStorageInvalidator',
-    //                ],
-    //            ]
-    //        ]
-    //    ]
```
kulla, line 89: Thanks! I noticed that `./CacheInvalidator/src/Invalidator/NavigationStorageInvalidator.php` and other here mentioned files are still in the repo without being called. I guess they can be deleted...

kulla (commented):
First review. I'll continue tomorrow. So far is the bug when subjects contain a digit the only blocker I have for bringing this into production.

packages/public/server/src/module/Common/src/Filter/Slugify.php
inyono: Since we don't have tests for most stuff in-place, I left everything I didn't need to change untouched. In this specific case, `Slugify` implements the `ZendFilterInterface` and this is why this is called via the public method `filter` instead. Don't ask me whether Zend does anything special when something implements the `ZendFilterInterface`.

packages/public/server/src/module/Common/src/Filter/Slugify.php
inyono: Handled mostly be the rest of the regexp. It probably still adds a minus to the beginning / end if there's something like space, but tbh, I don't care much about that. Might even be the case that we clean the title up a little bit anyways.

packages/public/server/src/module/Common/src/Filter/Slugify.php
inyono: Well, similar to the stuff above: since we don't have tests in place, this is always a little tricky since PHP (and Zend especially) doesn't always use the stuff in a way that a dead code detector recognizes it (e.g. `fooAction` in a controller is mentioned as `foo` in the `module.config.php`. Similarly, except in newer code, we refer to classes via a string (sometimes even a dynamic string using `__NAMEPACE__`)...)

packages/public/server/src/module/Alias/src/Controller/AliasController.php
inyono: Fix in https://github.com/serlo/serlo.org/pull/543/commits/8dc26d029a7ebc257abf62fd168ca46d40f33d16

packages/public/server/src/module/Alias/src/Controller/AliasController.php
```diff
@@ -22,71 +22,149 @@
...
         }
+    }
 
+    private function resolveAlias($alias)
+    {
+        if (preg_match('/(?<id>\d+)\//', $alias, $matches)) {
+            try {
+                $object = $this->uuidManager->getUuid($matches['id'], true);
+                $normalized = $this->normalizer->normalize($object);
+                return $this->getUrlOfNormalizedObject($normalized);
```
kulla, line 85:
```suggestion
                return $this->resolveUuid($matches['id']);
```

Why don't we use `resolveUuid()` here? To have a fallback "UUID not found, fall through to check if this is a legacy alias"?

packages/public/server/src/module/Common/src/Filter/Slugify.php
inyono: It seems to be used in a couple of places:
Screen Shot 2020-11-26 at 16 04 26


Though If I understand the documentation correctly, it just allows you to use some built-in helpers of Zend but doesn't matter much otherwise. So we could refactor them and don't use `Zend\Filter\FilterInterface` at all.

packages/public/server/src/module/Alias/src/Controller/AliasController.php
inyono: Yep. `resolveUuid` is only used for `/:id` paths (there we have to resolve via ID and don't want to look up an alias). `resolveAlias` is used for the remaining paths. Also note that the behavior for `/:id` vs. `/:some-alias` is different because `/:id` has the custom logic for injections (which didn't really work when using aliases. Since this isn't a supported use case anyways, I'd leave it like that for now).

packages/public/server/src/module/Alias/src/Controller/AliasController.php
inyono: We can do that when we need it somewhere else, too.

packages/public/server/src/module/Common/src/Filter/Slugify.php
kulla: I meant `Slugify` is only used at `SubjectHelper` -> Maybe the other places will help identify why `Zend\Filter\FilterInterface` is needed. I would guess that it can be added to a Form...

Issue #560, opened 2020-12-07
Listener is missing for creation of topic


Entkenntnis:
When you create a topic, the cache for the taxonomy parent is not updated. The legacy server needs to trigger an update for the API an the parent.

Ways to reproduce:

1. Create new topic in any taxonomy.
2. The topic is not shown in the new frontend and in the API

PR #561, opened 2020-12-08
server: also update parent on create term

Entkenntnis:
An idea for fixing #560


inyono:
This won't resolve the issue since the parent's cache is still not set. This changes only the behavior of the legacy system itself (that worked correctly so far to my understanding).

PR #570, opened 2021-01-03
Add hash to redirects when storing a revision

### LINKED ISSUE #553: Add hashes to redirects after saving and accepting revisions (for frontend)
For users who use the new frontend after a successful save should add a hash to the following redirect.

e.g. instead of 
`…/entity/repository/history/1681#revision-saved` redirect to `…/entity/repository/history/1681#revision-saved` 
or for admins or within the sandkasten: instead of `…/sinus-kosinus-und-tangens` redirect to `…/sinus-kosinus-und-tangens#revision-saved-accepted`

The same logic applies to accepting revisions:
here we could add a `#revision-accepted` to the url.

kulla:
Closes #553 

This PR implements (ping @elbotho ):

* Adds `#revision-saved` when a revision is saved but needs to be reviewed
* Adds `#revision-accepted` when a revision was successfully checkout out / accepted
* Adds `#revision-rejected` when a revision was successfully rejected
* Adds `#revision-saved-accepted` when a revision was successfully saved and accepted

### DIFF
--- a/packages/public/server/src/module/Entity/src/Controller/RepositoryController.php
+++ b/packages/public/server/src/module/Entity/src/Controller/RepositoryController.php
@@ -85,11 +85,12 @@ class RepositoryController extends AbstractController
         if ($this->getRequest()->isPost()) {
             $form->setData($this->getRequest()->getPost());
             if ($form->isValid()) {
-                return $this->handleAddRevisionPost(
+                $redirectUrl = $this->handleAddRevisionPost(
                     $entity,
                     $form->getData(),
                     true
                 );
+                return $this->redirect()->toUrl($redirectUrl);
             }
         }
 
@@ -437,24 +438,19 @@ class RepositoryController extends AbstractController
                 'Your revision has been saved and is available'
             );
             $route = 'entity/page';
+            $hash = '#revision-saved-accepted';
         } else {
             $successMessage = $translator->translate(
                 'Your revision has been saved, it will be available once it gets approved'
             );
             $route = 'entity/repository/history';
+            $hash = '#revision-saved';
         }
         $this->getEntityManager()->flush();
         $this->flashMessenger()->addSuccessMessage($successMessage);
 
-        if ($this->getRequest()->isXmlHttpRequest()) {
-            return $this->plugin('url')->fromRoute($route, [
-                'entity' => $entity->getId(),
-            ]);
-        }
-
-        return $this->redirect()->toRoute($route, [
-            'entity' => $entity->getId(),
-        ]);
+        $entityId = ['entity' => $entity->getId()];
+        return $this->plugin('url')->fromRoute($route, $entityId) . $hash;
     }
 
     public function checkoutAction()
@@ -477,9 +473,9 @@ class RepositoryController extends AbstractController
         );
         $this->getRepositoryManager()->flush();
 
-        return $this->redirect()->toRoute('entity/page', [
-            'entity' => $entity->getId(),
-        ]);
+        $entityId = ['entity' => $entity->getId()];
+        $url = $this->plugin('url')->fromRoute('entity/page', $entityId);
+        return $this->redirect()->toUrl($url . '#revision-accepted');
     }
 
     public function compareAction()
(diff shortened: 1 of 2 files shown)


elbotho:
👍 Thanks!

packages/public/server/src/module/Entity/src/Controller/RepositoryController.php
```diff
@@ -437,24 +438,19 @@ protected function handleAddRevisionPost(
...
-        if ($this->getRequest()->isXmlHttpRequest()) {
-            return $this->plugin('url')->fromRoute($route, [
-                'entity' => $entity->getId(),
-            ]);
-        }
-
-        return $this->redirect()->toRoute($route, [
-            'entity' => $entity->getId(),
-        ]);
+        $entityId = ['entity' => $entity->getId()];
```
inyono, line 452: Why is this an extra variable? Isn't needed and additionally confusing because `$entityId` is not the entity's id.

packages/public/server/src/module/Entity/src/Controller/RepositoryController.php
```diff
@@ -477,9 +473,9 @@ public function checkoutAction()
         );
         $this->getRepositoryManager()->flush();
 
-        return $this->redirect()->toRoute('entity/page', [
-            'entity' => $entity->getId(),
-        ]);
+        $entityId = ['entity' => $entity->getId()];
```
inyono, line 476: Same here.

packages/public/server/src/module/Entity/src/Controller/RepositoryController.php
```diff
@@ -437,24 +438,19 @@ protected function handleAddRevisionPost(
...
             $route = 'entity/repository/history';
+            $hash = '#revision-saved';
         }
         $this->getEntityManager()->flush();
         $this->flashMessenger()->addSuccessMessage($successMessage);
 
-        if ($this->getRequest()->isXmlHttpRequest()) {
-            return $this->plugin('url')->fromRoute($route, [
-                'entity' => $entity->getId(),
-            ]);
```
inyono, line 452: You don't seem to handle `XMLHttpRequests` anymore. What's the reason for that? This was necessary for POST requests from JavasScript to the URL (e.g. Edtr.io).

packages/public/server/src/module/Entity/src/Controller/RepositoryController.php
```diff
@@ -85,11 +85,12 @@ public function addLegacyRevisionAction()
         if ($this->getRequest()->isPost()) {
             $form->setData($this->getRequest()->getPost());
             if ($form->isValid()) {
-                return $this->handleAddRevisionPost(
+                $redirectUrl = $this->handleAddRevisionPost(
```
inyono, line 88: Pretty sure that this won't work for Edtr.io stuff since that expected the redirect url in the response body (instead of redirecting directly), see: https://github.com/serlo/serlo.org/blob/974c9e5279d5e209015b2b05d5374d7c1dfc5f69/packages/public/client/src/editor/editor/index.tsx#L54-L62.

inyono (changes requested):

packages/public/server/src/module/Entity/src/Controller/RepositoryController.php
kulla: Before the function returned the url as a string if it was an XMLHttpRequest and a Zend Redirect Response if not. Now it only returns a string. For the old markdown editor the string is converted into a zend redirect at https://github.com/serlo/serlo.org/blob/c0739c6fb060583f58e5b874281b23df575f443f/packages/public/server/src/module/Entity/src/Controller/RepositoryController.php#L88-L93

packages/public/server/src/module/Entity/src/Controller/RepositoryController.php
kulla: This is for the old editor `function addLegacyRevisionAction()`. At https://github.com/serlo/serlo.org/blob/c0739c6fb060583f58e5b874281b23df575f443f/packages/public/server/src/module/Entity/src/Controller/RepositoryController.php#L132-L142 you have the code for edtr.io

kulla:
@inyono Code suggestions implemented. Ready for review.

packages/public/server/src/module/Entity/src/Controller/RepositoryController.php
inyono: Check, `addLegacyRevisionAction` is always a normal HttpRequest while `addRevisionAction` is always a XMLHttpRequest, so the check is no longer needed.

inyono (approved):

PR #572, opened 2021-01-04
server: Add /api/add-comment

kulla:
@inyono Ready for first feedback (since I have questions where I need you help :smile: ) My questions are:

### ~FormInterface is used as arguments~

The function https://github.com/serlo/serlo.org/blob/cf9c2df36a81a7ff23bb68d7f2e4d52a401b8854/packages/public/server/src/module/Discussion/src/DiscussionManager.php#L74-L99 uses `FormInterface` as an argument. I want to discuss what an appropriate type shall be for API mutations. In the current PR I have created a new comment with `new Comment()`, stored all necessary arguments and saved it with `$this->getObjectManager()->persist()` + `->flush()`. Is this a good way to create a new uuid?

### ~User is not logged in~

Since the api does not forward the session to the legacy serlo.org the current user is not logged in but given via the `userId` parameter. However the function https://github.com/serlo/serlo.org/blob/cf9c2df36a81a7ff23bb68d7f2e4d52a401b8854/packages/public/server/src/module/Discussion/src/DiscussionManager.php#L88 checks against the currently logged in user. How shall we implement it.

### What is missing:

- [x] Check whether user has rights to create a comment.
- [x] Check whether there is no comment to an comment to an comment
- [x] Add new event_log() entry

### DIFF
--- a/packages/public/server/__tests-e2e__/api.ts
+++ b/packages/public/server/__tests-e2e__/api.ts
@@ -33,86 +33,63 @@ describe('/api/alias/:alias', () => {
   })
 })
 
-describe('/api/add-comment', () => {
-  const exampleBody = {
-    title: null,
+const mutationEndpoints = {
+  '/api/thread/start-thread': {
+    title: 'A new thread',
     content: 'Hello World',
     objectId: 1855,
     userId: 10,
-  }
+  },
+  '/api/thread/comment-thread': {
+    content: 'Hello World',
+    objectId: 27778,
+    userId: 10,
+  },
+}
 
-  describe('adds a new comment', () => {
-    describe.each([
-      [
-        'when objectId is not a comment and title is a given',
-        { objectId: 1855, title: 'A new thread' },
-        getExpectedCreateThreadEvent,
-      ],
-      [
-        'when objectId is not a comment and title is null',
-        { objectId: 1855, title: null },
-        getExpectedCreateThreadEvent,
-      ],
-      [
-        'comments a new thread when objectId is a comment',
-        { objectId: 27778 },
-        getExpectedCreateCommentEvent,
-      ],
-    ])('%s', (_name, bodyDiff, getTargetEvent) => {
-      const body = { ...exampleBody, ...bodyDiff }
-      let comment: unknown
+describe('/api/thread/start-thread', () => {
+  const body = mutationEndpoints['/api/thread/start-thread']
 
-      beforeAll(async () => {
-        await fetchApi('/api/e2e-tests/set-up')
+  describe('starts a new thread', () => {
+    const expectedPayload = {
+      __typename: 'Comment',
+      alias: expect.any(String),
+      archived: false,
+      authorId: body.userId,
+      childrenIds: [],
+      date: expect.any(String),
+      id: expect.any(Number),
+      content: body.content,
+      parentId: body.objectId,
+      title: body.title,
+      trashed: false,
+    }
+    let comment: unknown
 
-        const response = await fetchApi('/api/add-comment', withJsonBody(body))
-        comment = await response.json()
-      })
+    beforeAll(async () => {
+      await fetchApi('/api/e2e-tests/set-up')
 
-      test('returns comment payload of newly created comment', async () => {
-        assertExpectedPayload(body, comment)
-      })
-
-      test('returns the same payload as /api/uuid/:id', async () => {
-        assertExpectedPayload(body, comment)
-        const response = await fetchApi(`/api/uuid/${comment.id}`)
-        expect(await response.json()).toEqual(comment)
-      })
-
-      test('creates a new event to the event log', async () => {
-        assertExpectedPayload(body, comment)
-        const response = await fetchApi('/api/e2e-tests/events-since-set-up')
-        const { events } = await response.json()
-
-        expect(events).toContainEqual(getTargetEvent(comment))
-      })
+      const init = withJsonBody(body)
+      const response = await fetchApi('/api/thread/start-thread', init)
+      comment = await response.json()
     })
 
-    function assertExpectedPayload(
-      body: Body,
-      comment: unknown
-    ): asserts comment is CommentPayload {
-      expect(comment).toEqual(getExpectedPayload(body))
-    }
+    test('returns comment payload of newly created comment', async () => {
+      expect(comment).toEqual(expectedPayload)
+    })
 
-    function getExpectedPayload(body: Body) {
-      return {
-        __typename: 'Comment',
-        alias: expect.any(String),
-        archived: false,
-        authorId: body.userId,
-        childrenIds: [],
-        date: expect.any(String),
-        id: expect.any(Number),
-        content: body.content,
-        parentId: body.objectId,
-        title: body.title,
-        trashed: false,
-      }
-    }
+    test('returns the same payload as /api/uuid/:id', async () => {
+      assertExpectEqual(comment, expectedPayload)
+      const response = await fetchApi(`/api/uuid/${comment.id}`)
+      expect(await response.json()).toEqual(comment)
+    })
 
-    function getExpectedCreateThreadEvent(comment: CommentPayload) {
-      return {
+    test('creates a new event to the event log', async () => {
+      assertExpectEqual(comment, expectedPayload)
+      const response = await fetchApi('/api/e2e-tests/events-since-set-up')
+      const { events } = await response.json()
+
+      expect(events).toContainEqual({
         id: expect.any(Number),
         instance: 'de',
         date: matchDate(comment.date),
@@ -120,11 +97,59 @@ describe('/api/add-comment', () => {
         __typename: 'CreateThreadNotificationEvent',
         actorId: comment.authorId,
         threadId: comment.id,
-      } as unknown
+      })
+    })
+  })
+
+  test('returns 400 when uuid is not commentable', async () => {
+    const init = withJsonBody({ ...body, objectId: 1 })
+    await assert400('/api/thread/start-thread', init)
+  })
+})
+
+describe('/api/thread/comment-thread', () => {
+  const body = mutationEndpoints['/api/thread/comment-thread']
+
+  describe('adds a new comment', () => {
+    let comment: unknown
+    const expectedPayload = {
+      __typename: 'Comment',
+      alias: expect.any(String),
+      archived: false,
+      authorId: body.userId,
+      childrenIds: [],
+      date: expect.any(String),
+      id: expect.any(Number),
+      content: body.content,
+      parentId: body.objectId,
+      title: null,
+      trashed: false,
     }
 
-    function getExpectedCreateCommentEvent(comment: CommentPayload) {
-      return {
+    beforeAll(async () => {
+      await fetchApi('/api/e2e-tests/set-up')
+
+      const init = withJsonBody(body)
+      const response = await fetchApi('/api/thread/comment-thread', init)
+      comment = await response.json()
+    })
+
+    test('returns comment payload of newly created comment', async () => {
+      expect(comment).toEqual(expectedPayload)
+    })
+
+    test('returns the same payload as /api/uuid/:id', async () => {
+      assertExpectEqual(comment, expectedPayload)
+      const response = await fetchApi(`/api/uuid/${comment.id}`)
+      expect(await response.json()).toEqual(comment)
+    })
+
+    test('creates a new event to the event log', async () => {
+      assertExpectEqual(comment, expectedPayload)
+      const response = await fetchApi('/api/e2e-tests/events-since-set-up')
+      const { events } = await response.json()
+
+      expect(events).toContainEqual({
         id: expect.any(Number),
         objectId: comment.id,
         instance: 'de',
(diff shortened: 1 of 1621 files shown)


inyono:
> Since the api does not forward the session to the legacy serlo.org the current user is not logged in but given via the `userId` parameter. However the function
> 
> https://github.com/serlo/serlo.org/blob/cf9c2df36a81a7ff23bb68d7f2e4d52a401b8854/packages/public/server/src/module/Discussion/src/DiscussionManager.php#L88
> 
> checks against the currently logged in user. How shall we implement it.

I'd create a new helper for that (you might want some different behavior there anyways since the current view helper also triggers a redirect to the login page when not authorized).



> In the current PR I have created a new comment with `new Comment()`, stored all necessary arguments and saved it with `$this->getObjectManager()->persist()` + `->flush()`. Is this a good way to create a new uuid?

IMHO we should still use the form since this already checks if everything is valid. So do something similar to `addRevision`, i.e. initialize the form, set the data and then check if everything is valid.

kulla:
> > 
> 
> I'd create a new helper for that (you might want some different behavior there anyways since the current view helper also triggers a redirect to the login page when not authorized).

Can you point me to the helper where this is currently implemented?

inyono:
> Can you point me to the helper where this is currently implemented?

- https://github.com/serlo/serlo.org/blob/master/packages/public/server/src/module/Authorization/config/module.config.php#L72
- https://raw.githubusercontent.com/serlo/serlo.org/master/packages/public/server/src/module/Authorization/src/Controller/Plugin/AssertGranted.php

packages/public/server/src/module/Api/config/module.config.php
```diff
@@ -198,6 +207,15 @@
...
                             ],
+                            'event' => [
+                                'type' => 'segment',
+                                'options' => [
+                                    'route' => '/last-event',
+                                    'defaults' => [
+                                        'action' => 'lastEvent',
+                                    ],
+                                ],
+                            ],
```
kulla, line 218: Fetches the last event -> it is used in the tests that a new event is created when a new comment is stored

kulla, line 218: Fetches the last event -> it is used in the tests that a new event is created when a new comment is stored

packages/public/server/src/module/Api/src/Controller/ApiController.php
```diff
@@ -69,6 +81,100 @@ public function aliasAction()
...
+                'author' => $user,
+                'instance' => $instance,
+                'title' => $data['title'],
+                'content' => $data['content'],
+                'csrf' => CsrfTokenContainer::getToken(),
+                'subscription' => [
+                    'subscribe' => true,
+                    'mailman' => true,
+                ],
+            ]);
```
kulla, line 171: Note to me: `$form->setData()` can only be called once (a new call overrides the old data).

packages/public/server/src/module/Api/src/Manager/NotificationApiManager.php
```diff
@@ -29,6 +31,8 @@ class NotificationApiManager
     /** @var Raven_Client */
     protected $sentry;
 
+    use ClassResolverAwareTrait, EntityManagerAwareTrait;
```
kulla, line 34: https://github.com/serlo/serlo.org/pull/508 will also use them

packages/public/server/src/module/Authentication/src/Service/AuthenticationService.php
```diff
@@ -128,6 +142,11 @@ public function hasIdentity()
         return $return;
     }
 
+    public function setIdentity(UserInterface $identity)
+    {
+        $this->identity = $identity;
+    }
```
kulla, line 148: This is used to simulate that `$identity` is authenticated in the current session.

kulla (commented):
@inyono I am ready for new feedback

packages/public/server/src/module/Authentication/src/Service/AuthenticationService.php
```diff
@@ -62,6 +63,11 @@ class AuthenticationService extends ZendAuthenticationService implements
      */
     protected $cookiePath = '/';
 
+    /**
+     * @var UserInterface
+     */
+    protected $identity = null;
+
```
inyono, line 70: As I said above: we probably don't want to use this `AuthenticationService` for that. Please create a new `AuthenticationService` that is only intended for the API where we have a better way to make sure that we don't introduce side effects. If feasible, we don't want to use an `AuthenticationService` at all (since this uses `Zend\Authentication\AuthenticationService` under the hood, where we again have no control over what's happening).

packages/public/server/src/module/Discussion/src/Controller/DiscussionController.php
```diff
@@ -105,7 +105,6 @@ public function commentAction()
             $form->setData(array_merge($this->params()->fromPost(), $data));
             if ($form->isValid()) {
                 $this->getDiscussionManager()->commentDiscussion($form);
-                $this->getDiscussionManager()->flush();
```
inyono, line 108: Why are those removed? The flushes are usually necessary (especially for the API stuff) since they make sure that the corresponding ids exists for upcoming events.

packages/public/server/src/module/Discussion/src/Controller/DiscussionController.php
```diff
@@ -167,7 +166,6 @@ public function startAction()
             $form->setData(array_merge($this->params()->fromPost(), $data));
             if ($form->isValid()) {
                 $this->getDiscussionManager()->startDiscussion($form);
-                $this->getDiscussionManager()->flush();
```
inyono, line 170: Same here.

inyono (changes requested):
Noted a couple of things. I think the major issue we should address before merging is the `AuthenticationService` stuff.

packages/public/server/__tests-e2e__/api.ts
```diff
@@ -19,28 +19,157 @@
...
+        parentId: 1855,
+        title: 'A new thread',
+        trashed: false,
+      })
+    })
+
+    test('returns the same payload as /api/uuid/:id', async () => {
+      const response = await fetchApi(`/api/uuid/${comment.id}`)
+      expect(await response.json()).toEqual(comment)
+    })
```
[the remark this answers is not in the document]
kulla: That is done at api.serlo.org. api.serlo.org uses also this endpoint for fetching the data.

packages/public/server/__tests-e2e__/api.ts
```diff
@@ -19,28 +19,157 @@
...
     test('returns null when user does not exist', async () => {
-      const data = await fetchPath('/api/alias/user/profile/not-existing')
-      expect(data).toBeNull()
+      const response = await fetchApi('/api/alias/user/profile/not-existing')
+      expect(await response.json()).toBeNull()
     })
   })
 })
 
+describe('/api/add-comment', () => {
```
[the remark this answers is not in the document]
kulla: Uff, no special treatment. I need to take care that `yarn mysql:dump` is never called after I run any e2e tests (actually it's the same issue with the other tests).

Possible solution: We can run `yarn mysql:rollback` after any e2e tests

packages/public/server/__tests-e2e__/api.ts
```diff
@@ -19,28 +19,157 @@
...
+    expect(response.status).toBe(400)
+  })
+
+  test('returns 400 when objectId does not belong to an uuid', async () => {
+    const body = { ...payload, objectId: 100000000 }
+    const response = await fetchApi('/api/add-comment', withJsonBody(body))
+    expect(response.status).toBe(400)
+  })
+
+  test('returns 400 when userId does not belong to a user', async () => {
```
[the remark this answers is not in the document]
kulla: Actually this can only happen when api.serlo.org has a bug since `userId = null` should result in an authentication error.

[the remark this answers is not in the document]
kulla: But yeah, I'll use 403...

packages/public/server/__tests-e2e__/api.ts
```diff
@@ -19,28 +19,157 @@
...
+function fetchApi(path: string, init?: RequestInit) {
+  return fetch(testingServerUrl + path, init)
+}
+
+function withJsonBody(body: unknown): RequestInit {
+  return withBody(JSON.stringify(body))
+}
+
+function withMalformedJson(): RequestInit {
+  return withBody('Malformed JSON')
```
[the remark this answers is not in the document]
kulla: The `"` is missing

packages/public/server/src/module/Api/src/Controller/ApiController.php
```diff
@@ -69,6 +81,100 @@ public function aliasAction()
...
+                false,
+                false
+            );
+        } catch (NotFoundException $exception) {
+            $this->getResponse()->setStatusCode(Response::STATUS_CODE_400);
+            return $this->response;
+        }
+
+        if (
+            $uuid instanceof EntityInterface ||
```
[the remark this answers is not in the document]
kulla: You can. The `else` part is not allowed

packages/public/server/src/module/Discussion/src/Controller/DiscussionController.php
kulla: I moved it into the `commentDiscussion()` function at https://github.com/serlo/serlo.org/blob/a64a5629c4a47867cbfd81bd8fe6a9492323f407/packages/public/server/src/module/Discussion/src/DiscussionManager.php#L85-L95

packages/public/server/src/module/Discussion/src/Controller/DiscussionController.php
kulla: Moved into `startDiscussion()`

packages/public/server/src/module/Api/src/Controller/AbstractApiController.php
```diff
@@ -53,4 +58,27 @@ protected function assertAuthorization()
...
+                isset($data[$key]) &&
+                $validators[$key]($data[$key]);
+        }, $keys);
+
+        if ($isValid) {
+            return $data;
+        } else {
+            throw new \TypeError();
+        }
+    }
```
kulla, line 83: I haven't found a better way to implement it with a form. I would suggest to do this in a follow up issue.

kulla (commented):
@inyono Ready for feedback / to be merged

packages/public/server/src/module/Api/src/Controller/AbstractApiController.php
inyono: Shouldn't this already be covered in your code anyways, e.g.:
```php
$data = $this->getRequestBody([
    'content' => 'is_string',
    'userId' => 'is_int',
    'threadId' => 'is_int',
]);

// ...

/** @var CommentForm */
$form = $this->getServiceLocator()->get(CommentForm::class);
// The form will validate the given data.
$form->setData([
    'parent' => $uuid,
    'author' => $user,
    'instance' => $uuid->getInstance(),
    'content' => $data['content'],
    'csrf' => CsrfTokenContainer::getToken(),
    'subscription' => [
        'subscribe' => true,
        'mailman' => true,
    ],
]);

```

packages/public/server/src/module/Authorization/src/Service/AuthorizationService.php
```diff
@@ -30,12 +30,15 @@ class AuthorizationService extends \ZfcRbac\Service\AuthorizationService
 {
     /**
      * @param \Rbac\Permission\PermissionInterface|string $permission
-     * @param null                                        $context
+     * @param mixed $context
+     * @param array $roles
      * @return bool
      */
-    public function isGranted($permission, $context = null)
+    public function isGranted($permission, $context = null, array $roles = null)
```
inyono, line 37: That looks like it will cause a runtime error.

Generally: you should be super careful when placing actual types in PHP (because PHP will throw at runtime if the types don't match). This is especially dangerous for legacy code where the comment might suggest that the type is always string but in practice it may be null. I'm not sure if PHP accepts `null` as array, but I'd triple-check that before adding the type here.

inyono (commented):
Looks okay, I'd check if the normal comment stuff still works (especially regarding the possible runtime error). Otherwise, this is probably okay for merge.

packages/public/server/src/module/Api/src/Controller/AbstractApiController.php
kulla: Unfortunately not. When `$data["title"]` does not exists PHP does not raise an exception but raises an uncatchable `undefined index` error. So a `isset($data["content"]) ? $data["content"] : ...` is needed. Also I haven't found an easy way so that the form makes a check for the `$title` whether it is a string or not.

kulla: Another problem is `$data["userId"]` -> With `(int) $data["userId"] ?? -1` a non empty string for `$data["userId"]` is converted into 1 and then aeneas is found in the database. So I ended with something like `$userId = isset($data["userId") && is_int($data["userId"]) ? $data["userId"] : -1`. I haven't found a way to avoid such a utility function.

inyono: Well, whatever. Let's leave it like that then since we'll move it to the database layer soon, anyway.

packages/public/server/src/module/Authorization/src/Service/AuthorizationService.php
kulla: Good catch! I removed the type hint there. There is `?array` for this situation, however our php version does not support it. I forgot to remove the type hint there.

kulla:
> Looks okay, I'd check if the normal comment stuff still works (especially regarding the possible runtime error). Otherwise, this is probably okay for merge.

I just checked: Locally I could start and comment threads

inyono (approved):

PR #580, opened 2021-01-15
server: Update api cache for parent when term is created

kulla:
Closes https://github.com/serlo/serlo.org/issues/560

### DIFF
--- a/packages/public/server/src/module/Taxonomy/src/Controller/TermController.php
+++ b/packages/public/server/src/module/Taxonomy/src/Controller/TermController.php
@@ -81,8 +81,14 @@ class TermController extends AbstractController
             ]);
             $form->setData($data);
             if ($form->isValid()) {
-                $this->getTaxonomyManager()->createTerm($form);
-                $this->getTaxonomyManager()->flush();
+                $term = $this->getTaxonomyManager()->createTerm($form);
+                if ($term->getParent()) {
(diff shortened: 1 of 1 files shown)


inyono (changes requested):
See #561: This won't resolve the issue since the parent's cache is still not set. This changes only the behavior of the legacy system itself (that worked correctly so far to my understanding).

Entkenntnis:
Isn't this the same fix as for the sorting-bug? It worked in that case as I remember. The flushing seems to help trigger cache updates. @kulla Have you tested it locally? Can you confirm that it works or not?

inyono:
Maybe I'm missing something. To me it seems like this code doesn't affect the API behavior at all (and only changes the "usual" behavior of `TaxonomyManager` which was already correct).

kulla:
> Maybe I'm missing something. To me it seems like this code doesn't affect the API behavior at all (and only changes the "usual" behavior of `TaxonomyManager` which was already correct).

It works (I have tried it locally). The function `flushUpdatesForTerm($term)` flushes and triggers an update event for `$term` for which there is an Listener in the API module. Maybe we can rename the function into `flushAndTriggerUpdate($term)` or we do the update triggering and the flush command in separate steps...

inyono:
Check. I'm looking forward to get rid of the whole event stuff 😅

inyono (approved):

PR #595, opened 2021-02-08
edtr-io: Add support for cohesive exercise groups

kulla:
@inyono A first feedback would be great since it is the first time I change an edtr-io plugin :smile:

Missing:

- [ ] Add translation strings
- [ ] When an cohesive exercise group got included via an injection it should be included with its subexercises and solutions

Current preview:

https://user-images.githubusercontent.com/1327215/107241226-11d14f00-6a2b-11eb-9437-f854e34824fc.mp4


packages/private/edtr-io/src/plugins/types/text-exercise-group.tsx
```diff
@@ -39,6 +39,7 @@ export const textExerciseGroupTypeState = entityType(
   {
     ...entity,
     content: editorContent(),
+    cohesive: boolean(false),
```
inyono, line 42: We should probably add a migration for that (even though this is probably handled correctly thanks to `undefined` being falsy).

inyono (commented):
LGTM

packages/private/edtr-io/src/plugins/types/text-exercise-group.tsx
inyono: Ah, yes, `deserialize.ts` is already failing because of this. You can handle that there.

kulla: How can I test this?

kulla: Ah, I see...

inyono: > How can I test this?

Generally: by checking what happens when you edit / render the state of existing Edtr.io documents. In this specific case there is probably no need for that since TypeScript already forbids you to pass an invalid state (types ftw). If this was more complicated, one approach might be to test that `deserialize` outputs the correct new state for existing Edtr.io documents. Alternatively: add a db migration that updates all existing Edtr.io documents to the new states.

kulla: Okay, fixed... :smile:

kulla: Actually the line `const isCohesive = cohesive.value ?? false` is the migration. Is this okay or shall I put this somewhere else?

packages/private/edtr-io/src/deserialize.ts
```diff
@@ -536,6 +536,7 @@ export function deserialize({
           content: serializeEditorState(
             toEdtr(deserializeEditorState(state.content))
           ),
+          cohesive: false,
```
inyono, line 539: Eh, that won't work. With this, any grouped text exercise will be considered non-cohesive (even those that you saved as cohesive).

Basically: `deserialize` is responsible to get raw data from the database and deserialize it to an Edtr.io document. For this specific case, it takes the `content` fields from the text-exercise-group and respective grouped-text-exercises. The returned `type-text-exercise-group` plugin is just a fake plugin used for the Editor, it won't be persisted itself. So: you either want to persist `cohesive` somewhere else (e.g. as a new field in the entity) or move it into the `content` of text-exercise-group.

inyono: If that brief explanation was too confusing, we can jump into a quick hangout instead.

packages/private/edtr-io/src/plugins/types/text-exercise-group.tsx
inyono: That line itself isn't doing anything because Edtr.io already makes sure that you get the correct type. So as a plugin developer, you can always assume that the types are as they are defined in your state (and what TypeScript tells you they are).

The migration itself should happen somewhere else. There are a couple of approaches for this:
- On database level > database migration
- On server-side > in `deserialize.ts`
- On client-side > by using the `migration` state helper Edtr.io exposes

kulla: Actually I got `cohesive.value === undefined` in my plugin. Thus the line `const isCohesive = cohesive.value ?? false`. Is this intended?

inyono: That shouldn't happen, so that sounds like a bug. So I'm very interested in a reproducable example here :)

kulla:
@inyono I am ready for new feedback ( and I have some questions, can we have a call?)

inyono:
What's the state of this PR? Ready for review?

kulla:
> What's the state of this PR? Ready for review?

Yes. Also ready for merge. I decided that saving cohesive as a parameter to entity is the best way currently.

inyono (approved):

Issue #605, opened 2021-03-26
Review notification listeners


inyono:
Finally got a reproduce case for the saving issues in staging:

1. Import staging database
2. Start API locally (and add it to `definitions.local.php`)
3. Try to save a new revision for /1855. Notice it takes quite a while. When logging, you'll notice the biggest factor are the notification listeners.

So we need to change that up. Some ideas:

- Atomic updates for notifications
- Combining multiple updates

#604 deactivates the notification listeners for now. So to test it locally, you need to revert those changes.

PR #608, opened 2021-03-27
server(api): add GraphQLMockService

inyono:
Adds `/graphql/list` and `/graphql/clear` to inspect the outgoing GraphQL requests locally. Behaves the same as `/mails/list` and `/mails/clear`:

- `/graphql/list` lists the timestamp, query, and variables for the any outgoing GraphQL request.
- `/graphql/clear` clears that list.

This should help debug issues like #605 and also enables the possibility of e2e tests for listeners.

### DIFF
--- a/packages/public/server/src/module/Api/src/Controller/GraphQLMockController.php
+++ b/packages/public/server/src/module/Api/src/Controller/GraphQLMockController.php
@@ -1,0 +1,68 @@
+<?php
+/**
+ * This file is part of Serlo.org.
+ *
+ * Copyright (c) 2013-2021 Serlo Education e.V.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License")
+ * you may not use this file except in compliance with the License
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * @copyright Copyright (c) 2013-2021 Serlo Education e.V.
+ * @license   http://www.apache.org/licenses/LICENSE-2.0 Apache License 2.0
+ * @link      https://github.com/serlo-org/serlo.org for the canonical source repository
+ */
+namespace Api\Controller;
+
+use Api\Service\AbstractGraphQLService;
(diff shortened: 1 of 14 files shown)


packages/public/server/src/module/Api/src/Controller/GraphQLMockController.php
```diff
@@ -0,0 +1,68 @@
...
+
+    public function __construct(AbstractGraphQLService $graphql)
+    {
+        if ($graphql instanceof GraphQLMockService) {
+            $this->active = true;
+            $this->graphql = $graphql;
+        }
+    }
+
+    public function listAction()
```
kulla, line 50: This would be only accessible in the local development environment, right?

packages/public/server/src/module/Api/src/Service/GraphQLMockService.php
```diff
@@ -0,0 +1,64 @@
...
+    {
+        $this->storage = $storage;
+        if (!$this->storage->hasItem($this->key)) {
+            $this->storage->setItem($this->key, []);
+        }
+    }
+
+    public function exec(string $query, array $variables)
+    {
+        $list = $this->storage->getItem($this->key);
```
kulla, line 46:
```suggestion
        $list = $this->get();
```

packages/public/server/src/module/Api/src/Controller/GraphQLMockController.php
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    public function clearAction()
+    {
+        if (!$this->active) {
+            $this->getResponse()->setStatusCode(404);
+            return false;
+        }
+        $this->graphql->clear();
+        return $this->redirect()->toRoute('api/list');
```
kulla, line 66: Does this work? I miss `graphql` somewhere in the path...

kulla (approved):
LGTM

packages/public/server/src/module/Api/src/Controller/GraphQLMockController.php
inyono: Yes, see https://github.com/serlo/serlo.org/blob/1076a56f86a3e7cbcceb721f48bb78bc960c0d3b/packages/public/server/src/module/Api/src/Controller/GraphQLMockController.php#L44-L47.

packages/public/server/src/module/Api/src/Controller/GraphQLMockController.php
inyono: That's the route name, not the path, see https://github.com/serlo/serlo.org/blob/1076a56f86a3e7cbcceb721f48bb78bc960c0d3b/packages/public/server/src/module/Api/config/module.config.php#L61-L93

Issue #620, opened 2021-04-27
Listener missing when course page order is changed


kulla:
Ordering of the course page ordering does not change API cache via a listener. Original Bug report:

![2021-04-27-161829_911x119_scrot](https://user-images.githubusercontent.com/1327215/116257289-36a8a900-a774-11eb-8bce-f1ccc2619c44.png)

PR #622, opened 2021-04-27
fix: Add listener when order of entity children is changed

kulla:
Fixes https://github.com/serlo/serlo.org/issues/620

### DIFF
--- a/packages/public/server/src/module/Api/src/Listener/LinkServiceListener.php
+++ b/packages/public/server/src/module/Api/src/Listener/LinkServiceListener.php
@@ -52,6 +52,13 @@ class LinkServiceListener extends AbstractListener
         $this->getApiManager()->setUuid($term);
     }
 
+    public function onSortChildren(Event $e)
+    {
+        /** @var UuidInterface $parent */
+        $parent = $e->getParam('parent');
+        $this->getApiManager()->removeUuid($parent->getId());
+    }
+
     public function attachShared(SharedEventManagerInterface $events)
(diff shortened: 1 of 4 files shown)


packages/public/server/src/module/Api/src/Listener/LinkServiceListener.php
```diff
@@ -52,6 +52,13 @@ public function onChange(Event $e)
         $this->getApiManager()->setUuid($term);
     }
 
+    public function onSortChildren(Event $e)
+    {
+        /** @var UuidInterface $parent */
+        $parent = $e->getParam('parent');
+        $this->getApiManager()->removeUuid($parent->getId());
```
kulla, line 59: `setUuid()` does not work here since `$parent` does not hold the new order of children yet. I guess the changes are not flushed and I haven't found a quick way to do this. @inyono If you want to have `setUuid()` at this place, let's open a follow up issue. However I want to discuss in the dev meeting whether using `removeCache` is the smarter solution...

inyono: > However I want to discuss in the dev meeting whether using `removeCache` is the smarter solution...

Yeah, let's discuss that. This change would at least go against one of the main principles of the listeners. Besides making sure that the cache is up-to-date (vs. a time-based cache for example), it should also improve request times (which would not be the case if we just remove the cache).

inyono (approved):

PR #623, opened 2021-04-27
fix: Update cache mutation queries

kulla:
Update for https://github.com/serlo/api.serlo.org/pull/325

### DIFF
--- a/packages/public/server/src/module/Api/src/Service/AbstractGraphQLService.php
+++ b/packages/public/server/src/module/Api/src/Service/AbstractGraphQLService.php
@@ -35,9 +35,13 @@ abstract class AbstractGraphQLService
     public function setCache($key, $value)
     {
         $query = <<<MUTATION
-          mutation _setCache(\$key: String!, \$value: JSON!) {
-            _setCache(key: \$key, value: \$value)
-          }
+            mutation _setCache(\$key: String!, \$value: JSON!) {
+                cache {
+                    _setCache(input: { key: \$key, value: \$value }) {
+                        success
+                    }
+                }
+            }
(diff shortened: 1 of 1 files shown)


packages/public/server/src/module/Api/src/Service/AbstractGraphQLService.php
```diff
@@ -36,8 +36,8 @@ public function setCache($key, $value)
     {
         $query = <<<MUTATION
             mutation _setCache(\$key: String!, \$value: JSON!) {
-                cache {
-                    _setCache(input: { key: \$key, value: \$value }) {
+                _cache {
+                    setCache(input: { key: \$key, value: \$value }) {
```
CarolinJaser, line 40: It should be `set(input: ...)`

packages/public/server/src/module/Api/src/Service/AbstractGraphQLService.php
```diff
@@ -53,8 +53,8 @@ public function removeCache($key)
     {
         $query = <<<MUTATION
             mutation _removeCache(\$key: String!) {
-                cache {
-                    _removeCache(input: { key: \$key }) {
+                _cache {
+                    removeCache(input: { key: \$key }) {
```
CarolinJaser, line 57: It should be `remove(input: ...)`

CarolinJaser (changes requested):
If this is to fit with the naming used in serlo/api.serlo.org#325 , the naming here should be revised.

kulla:
@CarolinJaser queries corrected...

CarolinJaser (approved):

Issue #637, opened 2021-05-18
"More"/"Weitere ungeprüfte Bearbeitungen" Button on de.serlo.org/entity/unrevised


kathongi:
Show all unrevised edits on de.serlo.org/entity/unrevised separated on multiple pages. 
1. first show only the oldest 100 unrevised edits on de.serlo.org/entity/unrevised
2. at the end of the page implement a button "More" or "Weitere ungeprüfte Bearbeitungen"
3. click on the new button: show the next 100 oldest unrevised edits on a new page 
4. continue showing the "more" button until all unrevised edits are visible

kathongi:
ping @kulla

kulla:
@kathongi: https://github.com/serlo/serlo.org/pull/639 has a fix. See: 
![2021-05-18-192502_694x339_scrot](https://user-images.githubusercontent.com/1327215/118697780-f20eab80-b80f-11eb-8a87-44ae93efd095.png) (Translation for German site ist missing) Alright?

kathongi:
Perfect Kulla!

PR #639, opened 2021-05-18
/entity/unrevised: Show button to show more revisions

kulla:
Closes https://github.com/serlo/serlo.org/issues/637: 
2021-05-18-192502_694x339_scrot


kulla:
@inyono Can you show me again what I need to do for i18n?

inyono:
- Update the translation file from source code (E.g. Poedit works well because it has PHP and Twig support). See my commit for the changes.
- Upload the translation file to Crowdin.
- Translate on Crowdin.
- Export translations (you can ping me for that because I have a script for that).

inyono (approved):
LGTM. Probably still makes sense to do the i18n stuff before merging.

kulla:
> * Update the translation file from source code (E.g. Poedit works well because it has PHP and Twig support). See my commit for the changes.

With which command? `yarn i18n:regenerate` hasn't worked for me...

> * Export translations (you can ping me for that because I have a script for that).

Can you add this script into this repo or is this not possible?!

inyono:
- `yarn i18n:regenerate` is only for the the client i18n. The server i18n has no CLI, unfortunately (because the twig-stuff has no good CLI). This is why we use Poedit for that.

- I could add that script to the repo but I would need to change it up so that it works on non-macOS and that it pulls the API token from somewhere.

Since we don't update i18n often and have a good solution for the new frontend, I never prioritized to make this process easier. If you feel that this is necessary, I can look into that.

kulla:
> Since we don't update i18n often and have a good solution for the new frontend, I never prioritized to make this process easier. If you feel that this is necessary, I can look into that.

Let's see whether we need it more often...

PR #647, opened 2021-06-11
Remove route /user/public

kulla:
Follow up of https://github.com/serlo/serlo.org-cloudflare-worker/pull/209 with which the route /user/public can never be reached.

### DIFF
--- a/packages/public/server/src/config/instance/navigation/default.php
+++ b/packages/public/server/src/config/instance/navigation/default.php
@@ -263,11 +263,6 @@ return [
                         'visible' => false,
                     ],
                     [
-                        'label' => t('Public profile'),
-                        'route' => 'user/public',
-                        'translate' => true,
-                    ],
-                    [
                         'label' => t('Edit profile'),
                         'route' => 'user/settings',
                         'icon' => 'wrench',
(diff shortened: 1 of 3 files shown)


inyono:
Does this also work correctly for logged-in users (that may be redirected to the legacy frontend at the moment)?

Entkenntnis:
"Eigenes Profil" directs to user profile directly, without /user/public, so frontend is fine

inyono:
> "Eigenes Profil" directs to user profile directly, without /user/public, so frontend is fine

I'm talking about legacy. I.e. does it work for users that are logged in and are still served the legacy system.

kulla:
> Does this also work correctly for logged-in users (that may be redirected to the legacy frontend at the moment)?

Yes. `/user/public` is a redirect to the profile route `/user/:id/:username` which is the same as `/user/me` without the navigation left with buttons like `change password`, `edit profile`, etc. This PR would delete the button `public profile` and thus it would be harder for users to get to their public profile view without navigation. However I think that's okay (since they see everything) and it is planned to serve the new design to all logged in users in two weeks anyway...

kulla:
(In the new design there is no distinction between a view on the user profile without and with a navigation -> both views are integrated in one)

inyono (approved):

PR #656, opened 2021-08-03
edtr-io: use react-virtual for text-exercise-groups

inyono:


inyono:
Want to check the production version, so deploying as v21. If that works as expected, I do the same at least for course pages.

PR #658, opened 2021-08-10
API: Remove cache instead of updating it

kulla:
Closes https://github.com/serlo/serlo.org/issues/657

### DIFF
--- a/packages/public/server/package.json
+++ b/packages/public/server/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@serlo/server",
-  "version": "18.0.11",
+  "version": "18.0.12",
   "private": true,
   "license": "Apache-2.0",
   "author": "Serlo Education e.V.",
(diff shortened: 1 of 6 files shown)


inyono (changes requested):
Shouldn't we be able to remove `setUuid`?

kulla:
> Shouldn't we be able to remove `setUuid`?

I only removed setUuid() at places where entities might be envolved. Here is the remaining changes: https://github.com/serlo/serlo.org/pull/659 Maybe we want to wait to see how the effects will be in production...

kulla:
@inyono What do you when we change `removeKey` to `updateKey` so that the key get's updated via SWR?

inyono:
> @inyono What do you when we change `removeKey` to `updateKey` so that the key get's updated via SWR?

We could do that, sure.

inyono (approved):

kulla:
@inyono This PR is now also live in production. When everything works fine we can also merge https://github.com/serlo/serlo.org/pull/659

PR #666, opened 2021-08-30
Refactoring of package deploy scripts

kulla:
Changes:

* Add support for beta versions -> See https://github.com/serlo/serlo.org/blob/1559c3dec13d00ae5113a4f3957e2bb1467697ed/packages/private/cloudflare/index.ts#L58-L65
* Replace package `requests` with `node-fetch`
* Use new token method instead of key + email, see https://www.tech-otaku.com/web-development/using-cloudflare-api-manage-dns-records/#24 -> **After the merge the `CF_EMAIL` can be deleted at https://github.com/serlo/serlo.org/settings/secrets/actions. It might be necessary to create a new token as well...**
* Remove unused function uploadWorker() + add some checks as well


kulla:
Seems to work :smile: 
2021-08-30-121357_888x135_scrot

package.json
```diff
@@ -18,7 +18,7 @@
     "c": "docker-compose run --rm server-composer",
     "deploy:images": "lerna run --stream deploy:image",
     "predeploy:packages": "lerna run --stream build --scope=@serlo/{client,static-assets} --include-filtered-dependencies",
-    "deploy:packages": "lerna run --stream deploy:package",
+    "deploy:packages": "cross-env DEPLOY=true lerna run --stream deploy:package",
```
inyono, line 21: Please don't, this is intentionally hard to deploy manually.

packages/private/cloudflare/index.ts
```diff
@@ -45,25 +46,31 @@ export async function publishPackage({
...
+
+    if (!prerelease) {
+      yield `${major}`
+      yield `${major}.${minor}`
+      yield `${major}.${minor}.${patch}`
+    } else {
+      for (let i = 1; i <= prerelease.length; i++) {
+        yield `${major}.${minor}.${patch}-${prerelease.slice(0, i).join('.')}`
+      }
+    }
```
inyono, line 73: ?

inyono: Why do we need this special syntax? Super hard to read. Could return an array instead.

packages/private/docker/index.ts
```diff
@@ -45,56 +64,50 @@ export function buildDockerImage({
...
+  if (!prerelease) {
+    yield 'latest'
+    yield `${major}`
+    yield `${major}.${minor}`
+    yield `${major}.${minor}.${patch}`
+  } else {
+    for (let i = 1; i <= prerelease.length; i++) {
+      yield `${major}.${minor}.${patch}-${prerelease.slice(0, i).join('.')}`
+    }
+  }
```
inyono, line 112: same

packages/private/docker/package.json
```diff
@@ -3,10 +3,5 @@
   "version": "0.0.0",
   "private": true,
   "dependencies": {},
-  "devDependencies": {
-    "@types/ramda": "^0.27.0",
-    "@types/semver": "^7.0.0",
-    "ramda": "^0.27.0",
-    "semver": "^7.0.0"
-  }
+  "devDependencies": {}
```
inyono, line 6: docker imports both `ramda` and `semver`

inyono (changes requested):

packages/private/cloudflare/index.ts
kulla: Have replaced it with a function returning an array. However coming from python the yield statement is quite natural to me :smile:

package.json
kulla: It is still hard now. Due to https://github.com/serlo/serlo.org/blob/e52cdf34ef3ec3007aa4804f4f8b0e8692cdf5a2/packages/private/cloudflare/index.ts#L109-L111 one gets an error when `CF_KEY` is not set. So nothing happens when one accidentally executes `yarn deploy:packages`. So only difference is that now the error message is more meaningful (I always wondered why the script said to me that it skips the deployment until I remembered that I have to set env variables...). Also with something like `CF_KEY=foo yarn deploy:packages` one gets the error `@serlo/static-assets: ✖  fatal     Error: CF: error getting "serlo-org-static-assets@2.0.0": {"success":false,"errors":[{"code":10000,"message":"Authentication error"}]} ` so without a valid token nobody can execute the script and we are the only ones who can create tokens for cloudflare...

kulla: With https://github.com/serlo/serlo.org/blob/e52cdf34ef3ec3007aa4804f4f8b0e8692cdf5a2/packages/private/cloudflare/index.ts#L109-L111 it is okay to delete the necessity of `DEPLOY=true`. It seems that it is just a safety against falsely executing `yarn deploy:images`.

packages/private/docker/package.json
kulla: I have moved them to the root `package.json`. Based on https://github.com/lerna/lerna#common-devdependencies I assume that collecting the devDependencies in one the repo is the way to go so that updating the dev dependencies is easier. See for example https://github.com/LittleBreak/lerna-best-practices -> there are no devDependencies in the child projects only on root level...

kulla: Anyhow: I do not mind to move the devDependencies back to the projects using them

package.json
inyono: IMHO: either remove `DEPLOY` completely or don't set it by default.

packages/private/docker/package.json
inyono: Hm... Not a big fan tbh, since it makes it harder to see where dependencies are used. But your call :)

package.json
kulla: Removed `DEPLOY` in https://github.com/serlo/serlo.org/pull/666/commits/728cf3c93d840cec50f0af2c3a08462de3e97fa2

kulla:
> See two remaining discussions.

@inyono Okay, ready to be merged

PR #672, opened 2021-09-01
docker: fix getTargetVersions

inyono:


inyono:
Already part of main.

PR #679, opened 2021-09-15
New approach to handle focus in equastions plugin

kulla:
**NOT READY TO BE MERGED**

This is a new approach to handle the focus in the equations plugin (see synopsis in https://serlo.slack.com/archives/CEB781NCU/p1631724649001000). Instead of intercepting the focus handling of the browser we use the `onFocus` events to track the current focus. The only necessarity is that only those HTML elements have `tabIndex={0}` which should be focusable via pressing `tab` or via mouse click. With `Shift+Tab` it works fine. With `tab` it only does not work because the elements of the tooltip of the math editor is focusable and the cycle through these controls first before we go to the next element. Changing this should work fine, though (edtr.io needs to be changed though for this)

### DIFF
--- a/packages/private/edtr-io/src/plugins/equations/editor.tsx
+++ b/packages/private/edtr-io/src/plugins/equations/editor.tsx
@@ -114,11 +114,13 @@ export function EquationsEditor(props: EquationsProps) {
   return (
     <HotKeys
       allowChanges
-      keyMap={{
+      keyMap={
+        /*{
         FOCUS_NEXT_OR_INSERT: 'tab',
         FOCUS_PREVIOUS: 'shift+tab',
         INSERT: 'return',
-      }}
+      }*/ {}
+      }
       handlers={{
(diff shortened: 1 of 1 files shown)


packages/private/edtr-io/src/plugins/equations/editor.tsx
```diff
@@ -319,7 +339,10 @@ function StepEditor(props: StepEditorProps) {
   return (
     <>
       <LeftTd
-        onClick={() => gridFocus.setFocus({ row, column: StepSegment.Left })}
+        tabIndex={
+          gridFocus.isFocused({ row, column: StepSegment.Left }) ? -1 : 0
```
inyono, line 343: That looks like a big anti-pattern. I'm pretty sure that you shouldn't modify `tabIndex` dynamically.

kulla:
Was just a test -> not necessary any more...

PR #682, opened 2021-09-21
Downgrade client on server -> release 19.0.2

kulla:


kulla:
@inyono Just noticed now that I could also change the `serlo-org-client@22` key in CF :see_no_evil: Okay, it's quite late...

PR #683, opened 2021-09-23
Add migration for first-explanation of equations

kulla:
@inyono Please have a detailed look since it is my first DB migration...

### DIFF
--- a/packages/migrations/src/20210923155000-add-first-explanation-to-equation.ts
+++ b/packages/migrations/src/20210923155000-add-first-explanation-to-equation.ts
@@ -19,7 +19,16 @@
  * @license   http://www.apache.org/licenses/LICENSE-2.0 Apache License 2.0
  * @link      https://github.com/serlo-org/serlo.org for the canonical source repository
  */
-export * from './clear-dead-uuids'
-export * from './create-migration'
-export * from './database'
-export * from './edtr-io'
+import { createEdtrIoMigration, updatePlugins } from './utils'
+
+createEdtrIoMigration({
+  exports,
+  migrateState: updatePlugins({
+    equations: ({ state }) => {
+      if (typeof state !== 'object' || state === null)
+        throw new Error('Illegal equation state')
+
+      return { firstExplanation: { plugin: 'text' }, ...state }
+    },
+  }),
+})
--- a/packages/migrations/src/utils/create-migration.ts
+++ b/packages/migrations/src/utils/create-migration.ts
@@ -19,6 +19,7 @@
  * @license   http://www.apache.org/licenses/LICENSE-2.0 Apache License 2.0
  * @link      https://github.com/serlo-org/serlo.org for the canonical source repository
  */
+import { isPlugin } from './edtr-io'
 import { CallbackBasedDatabase, createDatabase, Database } from './database'
 
 export function createMigration(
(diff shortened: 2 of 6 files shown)


packages/migrations/src/20210923155000-add-first-explanation-to-equation.ts
```diff
@@ -0,0 +1,34 @@
...
+import { createEdtrIoMigration, updatePlugins } from './utils'
+
+createEdtrIoMigration({
+  exports,
+  migrateState: updatePlugins({
+    equations: ({ state }) => {
+      if (typeof state !== 'object' || state === null)
+        throw new Error('Illegal equation state')
+
+      return { firstExplanation: { plugin: 'text' }, ...state }
```
inyono, line 31: This isn't good enough, is it? Don't we need to move the explanations "up"?

inyono: Also: did you test this with equations that already have a `firstExplanation`? Code looks fine but just to make sure.

packages/migrations/src/20210923155000-add-first-explanation-to-equation.ts
```diff
@@ -0,0 +1,34 @@
...
+ * @license   http://www.apache.org/licenses/LICENSE-2.0 Apache License 2.0
+ * @link      https://github.com/serlo-org/serlo.org for the canonical source repository
+ */
+import { createEdtrIoMigration, updatePlugins } from './utils'
+
+createEdtrIoMigration({
+  exports,
+  migrateState: updatePlugins({
+    equations: ({ state }) => {
+      if (typeof state !== 'object' || state === null)
```
inyono, line 28: Do we need that check?

packages/migrations/src/utils/edtr-io.ts
```diff
@@ -0,0 +1,69 @@
...
+      transformState: Transformation
+    }) => unknown
+  }
+): Transformation {
+  function transformState(value: unknown): unknown {
+    if (isPlugin(value)) {
+      const { plugin, state } = value
+      const transformFunc = transformations[plugin]
+
+      if (transformFunc !== undefined) {
```
inyono, line 37: Better: `typeof transformFunc === 'function'`

packages/migrations/src/utils/edtr-io.ts
```diff
@@ -0,0 +1,69 @@
...
+      }
+    }
+
+    if (Array.isArray(value)) {
+      return value.map(transformState)
+    }
+
+    if (typeof value === 'object' && value !== null) {
+      return R.mapObjIndexed(transformState, value)
+    }
```
inyono, line 48: That looks a little bit scary. Why don't you do the thing we did for the other migrations, i.e. actually going over the plugins where we know that they should have children? This will fail if we use the key `plugin` in some plugin's state or substate.

inyono: Also: does this even work if `state = undefined`?

packages/migrations/src/utils/create-migration.ts
```diff
@@ -58,4 +59,62 @@ export function createMigration(
...
+        SELECT erf.id, erf.value, er.id as revisionId
+        FROM entity_revision_field erf
+          LEFT JOIN entity_revision er on erf.entity_revision_id = er.id
+          LEFT JOIN entity e on er.repository_id = e.id
+          LEFT JOIN type entity_type on entity_type.id = e.type_id
+        WHERE erf.field = 'content'
+          AND entity_type.name IN ("article", "course-page",
+                                   "event", "grouped-text-excercise",
+                                   "text-exercise", "text-exercise-group",
+                                   "text-solution")
```
inyono, line 87: Why do we need only those types?

inyono (changes requested):

packages/migrations/src/20210923155000-add-first-explanation-to-equation.ts
kulla: Yeah, after a conversation with you I have add `firstExplanation` to avoid a breaking change in the plugin state (my understanding after our conversation was that then I would not need a database migration). Now we need a database migration anyway so I nothing gained from this :see_no_evil:

kulla: Since I have worked on this issue already for a long time I would like to have it like this and to move the migration to another time...

packages/migrations/src/20210923155000-add-first-explanation-to-equation.ts
kulla: We only need the check so that the typescript compiler is happy since state has the type `unknown`

packages/migrations/src/utils/create-migration.ts
kulla: The other types like `video` do not have edtr-io states, right?

packages/migrations/src/utils/edtr-io.ts
kulla: The beauty of it is that the code works regardless of the plugins we currently. Besides no state should have a `plugin` + `state` property...

packages/migrations/src/utils/create-migration.ts
inyono: They do:
```sql
SELECT DISTINCT t.name
FROM entity_revision_field erf
         JOIN entity_revision er on erf.entity_revision_id = er.id
         JOIN entity e on er.repository_id = e.id
         JOIN type t on e.type_id = t.id
WHERE erf.value LIKE '%"plugin"%'
```

packages/migrations/src/utils/edtr-io.ts
inyono: The problem with it is that it works on implicit assumptions. Especially for third-party code, the assumption might not hold. For example: are you sure that Slate's state doesn't include "plugin" + "state"?

Since we want to go over the migration process anyways and we move Edtr.io to the frontend soon-ish, I'm fine with merging this. We should think about another long-term solution, though.

PR #684, opened 2021-09-23
Equation improvements: Add mode for transform terms

kulla:


kulla:
@inyono Ready for review

Issue #348, opened 2020-03-29
Improvements of Input-exercises


yolo-bird:
- [ ] After choosing the input exercise
![Screenshot_2020-03-29 Serlo – Die freie Lernplattform(1)](https://user-images.githubusercontent.com/46370375/77859293-cf36c780-7208-11ea-9446-6a2f00960846.png)
the author should land directly in the field where he can add the solutions 
![Screenshot_2020-03-29 Serlo – Die freie Lernplattform](https://user-images.githubusercontent.com/46370375/77859315-f1304a00-7208-11ea-8ec1-eaa71a1975fb.png)
This reduces the number of clicks by one. 
- [ ] Currently the choice of which type the input exercise will be is hidden in the extended options
![Screenshot_2020-03-29 Ein Eingabefeld verwenden(1)](https://user-images.githubusercontent.com/46370375/77859430-99461300-7209-11ea-8d99-6eda6892fe0c.png)
We want the corresponding drop down menu to be between "activate" and the "first solution" 
![Screenshot_2020-03-29 Ein Eingabefeld verwenden](https://user-images.githubusercontent.com/46370375/77859480-f9d55000-7209-11ea-93ed-77f8458bf078.png)

- [ ] Rename the type of input-exercises
Zahl -> Zahl (eindeutiges Ergebnis: 0,5 \noteq 1/2 \noteq 2/4
Ausdruck -> Ausdruck (äquivalentes Ergebnis: 0,5= 1/2 =2/4)
Text -> Text

yolo-bird:
Moved to https://docs.google.com/document/d/1rSkFo2JtsSwHHsbosQcgAN4EzeePXLnqi_eCFpUM3u4/edit#

PR #687, opened 2021-09-30
sc-mc: Change some i18n strings

kulla:
See https://github.com/serlo/serlo.org/issues/348


inyono (changes requested):
Other language files need to be updated as well.

inyono (approved):

PR #688, opened 2021-09-30
exercise: Autofocus sc/mc/input after it is created

kulla:
See https://github.com/serlo/serlo.org/issues/348

@inyono It feels quite hacky this solution. Can we talk how we can add a solution to edtr.io to this problem?

### DIFF
--- a/packages/private/edtr-io/src/plugins/exercise.tsx
+++ b/packages/private/edtr-io/src/plugins/exercise.tsx
@@ -29,7 +29,7 @@ import {
   object,
   optional,
 } from '@edtr-io/plugin'
-import { getDocument } from '@edtr-io/store'
+import { focus, getDocument } from '@edtr-io/store'
 import { Icon, faRandom, faTrashAlt } from '@edtr-io/ui'
 import { useI18n, I18n } from '@serlo/i18n'
 import * as React from 'react'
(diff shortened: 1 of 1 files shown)


packages/private/edtr-io/src/plugins/exercise.tsx
```diff
@@ -120,6 +120,13 @@ function ExerciseEditor({ editable, state }: ExerciseProps) {
   const store = useScopedStore()
   const { content, interactive } = state
   const [showOptions, setShowOptions] = React.useState(false)
+  const [focusInteractive, setFocusInteractive] = React.useState(false)
```
inyono, line 123: - Why do we need a state?
- Why is the state never reseted after having been set to `true` once?

packages/private/edtr-io/src/plugins/exercise.tsx
```diff
@@ -120,6 +120,13 @@ function ExerciseEditor({ editable, state }: ExerciseProps) {
...
+
+  React.useEffect(() => {
+    if (interactive.defined && focusInteractive) {
+      store.dispatch(focus(interactive.id))
+    }
+  }, [focusInteractive])
```
inyono, line 129: Why does this not depend on `interactive.defined` and `interactive.id`?

inyono (changes requested):

inyono:
> It feels quite hacky this solution

Looks like a bug in the autofocus code (which we wanted to replace at some point anyway).

kulla:
@inyono Bug is fixed. Ready to be merged.

inyono (approved):

PR #689, opened 2021-10-06
Add table plugin

kulla:


kulla:
Changes copied to https://github.com/serlo/frontend/pull/1445 so that we can implement it directly in the new design.

Issue #651, opened 2021-07-14
Content API parameters in Search Results


elbotho:
Parameters like `?hideTopbar` and `?contentOnly` end up in google search results despite the `canonical` tag being set correctly.
Don't now why this happens, but it prevents the new frontend from showing and is confusing (since there is no menu).

elbotho:
Should we just set noindex to all legacy pages?

PR #694, opened 2021-10-31
noindex for legacy

elbotho:
for #651
any objections?
is that a bad idea? :D

### DIFF
--- a/packages/public/server/src/module/Normalizer/src/View/Helper/Normalize.php
+++ b/packages/public/server/src/module/Normalizer/src/View/Helper/Normalize.php
@@ -70,7 +70,7 @@ class Normalize extends AbstractHelper
         $meta->appendName('content_type', $type);
         $meta->appendName('description', $this->getMetaDescription($object));
         $meta->appendName('keywords', implode(', ', $keywords));
-        $meta->appendName('robots', $robots);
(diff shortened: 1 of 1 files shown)


kulla:
Lets talk about this tomorrow in the dev meeting...

kulla (approved):
Okay as promised I have looked to other solutions:

* https://www.gsqi.com/marketing-blog/google-ignore-rel-canonical-different-content/ explains why it happens and suggests the solution with `noindex, follow`

kulla:
I accidentally sent the comment too early. Here are the missing links:

kulla:
I accidentally sent the comment too early. Here are the missing links: 

* https://developers.google.com/search/docs/advanced/crawling/consolidate-duplicate-urls gives a good overview especially the video https://youtu.be/gEWkYTPSEjs
* Also https://i83.uk/google-appears-good-at-ignoring-canonical-tags-on-non-equivalent-pages/ shows that Google might store two urls as different records when though there is a canonical link between them

What happens: Google thinks that the legacy and the frontend version are different site due to the different site structure even though there is a canonical link between them. (The source of google states that the two site need to be almost the same). Thus this is the best solution until we migrate the content api to the new frontend

elbotho:
thanks @kulla !

PR #695, opened 2021-11-01
Create table deleted_bots

kulla:
After talking to @simonkoehl we noticed that we need to store deleted
bots since we want to remove them also from the newsletter later in the
futute. This PR creates a seperate table where we can store the data.

### DIFF
--- a/packages/migrations/src/202111012023800-create-table-deleted-bots.ts
+++ b/packages/migrations/src/202111012023800-create-table-deleted-bots.ts
@@ -1,0 +1,32 @@
+/**
+ * This file is part of Serlo.org.
+ *
+ * Copyright (c) 2013-2021 Serlo Education e.V.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License")
+ * you may not use this file except in compliance with the License
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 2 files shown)


packages/migrations/src/202111012023800-create-table-deleted-bots.ts
```diff
@@ -0,0 +1,32 @@
...
+import { createMigration } from './utils'
+
+createMigration(exports, {
+  up: async (db) => {
+    await db.runSql(`CREATE TABLE IF NOT EXISTS deleted_bots LIKE user`)
+  },
+  // TODO: @inyono Do I need a down-type?!
+  /* down: async (db) => {
+    await db.runSql(`INSERT INTO type (name) VALUES ('text-hint')`)
+  },*/
```
kulla, line 31: @inyono What is this down-command for? A possible undo for the migration?!

simonkoehl: thanks, @kulla !

inyono (changes requested):
Instead of persisting the deleted bots, why don't we directly remove them from the newsletter via Mailchimp API https://mailchimp.com/developer/marketing/api/list-members/delete-list-member/? Besides the id of the list, it only needs the (hashed) email address of the subscriber.

kulla:
As discussed we will implement another solution...

Issue #720, opened 2022-05-17
Add migration for image captions


kulla:
With https://github.com/serlo/frontend/pull/1651 we have added image captions. We need to add a DB migration adding image captions to all image plugins otherwise each time a course or exercise with images in the course-page / solution is added there is a changed edit in the review (even though the content itself hasn't changed)

PR #721, opened 2022-05-17
Add DB migration for image plugins

kulla:
Fixes https://github.com/serlo/serlo.org/issues/720


elbotho (approved):
tbh. I think it would be a lot nicer if edtr would save an empty value as `undefined`,`null` or `''` instead of ```{ plugin: 'text', state: [{ type: 'p', children: [{}] }] }```.

But that's a different topic. 
Code looks good, but I think I lack some knowledge to actually review this.

PR #732, opened 2022-07-01
api: Use "keys" instead of "key" in removeCache()

kulla:
See https://github.com/serlo/api.serlo.org/issues/687


inyono (changes requested):
As said in https://github.com/serlo/api.serlo.org/pull/688, this change should not be necessary.

PR #735, opened 2022-09-01
Migrate markdown table to new serlo table

kulla:


hugotiburtino:
@kulla do you think we can still  use this PR?

kulla:
Yes but by copy and pasting the code of this pr

